PR #2, opened 2018-11-19
Made sure node encoding strategy is considered for values inside unkeyed containers

regexident:
This fixes a bug that would make `XMLEncoder` ignore the node-encoding strategy for values contained in unkeyed collections.

Given a set of types like this …

```swift
struct Foo: Codable {
  var bar = Bar()
  var unkeyedBars = [Bar()]
  var keyedBars = ["foo": Bar()]
}

struct Bar: Codable {
  let value = "baz"
  
  static func nodeEncoding(forKey codingKey: CodingKey) -> XMLEncoder.NodeEncoding {
    return .attribute
  }
}
```

… and an encoder configured like this …

```swift
let encoder = XMLEncoder()
encoder.outputFormatting = .prettyPrinted
encoder.keyEncodingStrategy = .convertToSnakeCase
encoder.nodeEncodingStrategy = .custom { codableType, encoder in
  guard let barType = codableType as? Bar.Type else {
    return { _ in return .default }
  }
  return barType.nodeEncoding(forKey:)
}
```

… produces output that looks like this …

```xml
<foo>
    <bar value="baz" />
    <unkeyed_bars>
        <value>baz</value>
    </unkeyed_bars>
    <keyed_bars>
        <foo value="baz" />
    </keyed_bars>
</foo>
```

… while the correct output would look like this …

```xml
<foo>
    <unkeyed_bars value="baz" />
    <bar value="baz" />
    <keyed_bars>
        <foo value="baz" />
    </keyed_bars>
</foo>
```

### DIFF
--- a/Sources/XMLCoder/Encoder/XMLEncoder.swift
+++ b/Sources/XMLCoder/Encoder/XMLEncoder.swift
@@ -796,7 +796,15 @@ fileprivate struct _XMLUnkeyedEncodingContainer : UnkeyedEncodingContainer {
     
     public mutating func encode<T : Encodable>(_ value: T) throws {
         self.encoder.codingPath.append(_XMLKey(index: self.count))
-        defer { self.encoder.codingPath.removeLast() }
+        let nodeEncodings = self.encoder.options.nodeEncodingStrategy.nodeEncodings(
+            forType: T.self,
+            with: self.encoder
+        )
+        self.encoder.nodeEncodings.append(nodeEncodings)
+        defer {
+            let _ = self.encoder.nodeEncodings.removeLast()
+            self.encoder.codingPath.removeLast()
+        }
         self.container.add(try self.encoder.box(value))
     }
     
--- a/Tests/XMLCoderTests/NodeEncodingStrategyTests.swift
+++ b/Tests/XMLCoderTests/NodeEncodingStrategyTests.swift
@@ -1,0 +1,192 @@
+import XCTest
+@testable import XMLCoder
+
+class NodeEncodingStrategyTests: XCTestCase {
+    fileprivate struct SingleContainer: Encodable {
+        let element: Element
+        
+        enum CodingKeys: String, CodingKey {
+            case element = "element"
+        }
+    }
+    
+    fileprivate struct KeyedContainer: Encodable {
+        let elements: [String: Element]
+        
+        enum CodingKeys: String, CodingKey {
+            case elements = "element"
+        }
+    }
+    
+    fileprivate struct UnkeyedContainer: Encodable {
+        let elements: [Element]
+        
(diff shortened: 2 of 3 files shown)


Sources/XMLCoder/Encoder/XMLEncoder.swift
```diff
@@ -796,7 +796,16 @@ fileprivate struct _XMLUnkeyedEncodingContainer : UnkeyedEncodingContainer {
     
     public mutating func encode<T : Encodable>(_ value: T) throws {
         self.encoder.codingPath.append(_XMLKey(index: self.count))
-        defer { self.encoder.codingPath.removeLast() }
+        print("unkeyed type: \(type(of: value))")
```
MaxDesiatov: I guess this was added for debugging purposes and could be removed before merge?

MaxDesiatov:
Thank you @regexident, I'm pretty sure that codecov will pass after I merge #1 that gives us correct test coverage. In general I think that the library shouldn't use `print` for logging. Currently there is only one use of `print` and I expect even that one to go away once #3 is fixed. Long term it might make sense to implement some general logging settings available to users of `XMLCoder`.

MaxDesiatov:
Thanks for the quick update @regexident, would you mind also adding a separate test case with inline XML similar to what is done [here](https://github.com/MaxDesiatov/XMLCoder/blob/master/Tests/XMLCoderTests/XMLCoderTests.swift)? Looks like you already have test XML and Swift code that reproduce this.  

This new test should probably be added in a new separate file, as `XMLCoderTests` file is renamed in #1, which would cause conflicts if new tests were added to that existing test case.

regexident:
I actually did try adding test cases for this, alas `XMLEncoder` has non-deterministic node ordering (as doe `JSONEncoder` for that matter), due to using unordered dictionaries as internal storage, which these days make use of randomly salted seeds.

As such I'm somewhat unsure how to test this properly, other than e.g. doing `xml.contains("key=\"value\"")` for each expected key on the resulting XML string.

MaxDesiatov:
Wouldn't it be possible to test this with a single node? I hope the bug is still reproducible in that case. I guess we would always get the same result for a single node even with random ordering. So we only need to test that we don't get this result

```
<foo>
    <unkeyed_bars>
        <value>baz</value>
    </unkeyed_bars>
</foo>
```

instead of the correct one

```
<foo>
    <unkeyed_bars value="baz" />
</foo>
```

regexident:
That should work, yes.

For what it's worth I'm currently hacking on a refactor that would enable order-preservation, as XML often times expects elements to appear in a certain order. This should then also make testing much easier.

MaxDesiatov:
great stuff, thanks again!

regexident:
> That should work, yes.

Narrator: "It did."

PR #4, opened 2018-11-20
Moved `_XML…EncodingContainer` into their own files, matching decoder

regexident:
The `XMLEncoder.swift` is far too large as it is right now.

I thus moved `_XMLUnkeyedEncodingContainer` and `_XMLKeyedEncodingContainer` into their own respective files.

Note: The `Decoder` directory already does the same.


regexident:
🤔 this didn't change any logic. Just moving stuff around and changing some stuff from `fileprivate` to `internal`.

MaxDesiatov:
LGTM, again hoping that merge of #1 will make codecov checks pass and then this one's good to merge.

PR #9, opened 2018-11-22
Add NodeEncodingStrategies typelias as deprecated

MaxDesiatov:
This should prevent a source-breaking change and allows next XMLCoder release only with a minor version bump.


regexident:
👌🏻

PR #11, opened 2018-12-02
Honor .sortedKeys

qmoya:
I realized that the `.sortedKeys` formatting option was being ignored. Sorting keys is a convenient thing to have, particularly when testing—else, comparing against strings breaks all the time.

This PR wants to fix that.

It also adds a `.swiftformat` file. Its contents were inferred automatically by [SwiftFormat](https://www.github.com/nicklockwood/SwiftFormat) itself. I just wanted my contribution to be consistent with the project’s _de facto_ style, but I think all projects benefit from explicit formatting rules—thus this addition.

### DIFF
--- a/Sources/XMLCoder/Decoder/DecodingErrorExtension.swift
+++ b/Sources/XMLCoder/Decoder/DecodingErrorExtension.swift
@@ -23,7 +23,7 @@ internal extension DecodingError {
         let description = "Expected to decode \(expectation) but found \(_typeDescription(of: reality)) instead."
         return .typeMismatch(expectation, Context(codingPath: path, debugDescription: description))
     }
-
+    
     /// Returns a description of the type of `value` appropriate for an error message.
     ///
     /// - parameter value: The value whose type to describe.
@@ -38,12 +38,10 @@ internal extension DecodingError {
             return "a string/data"
         } else if value is [Any] {
             return "an array"
-        } else if value is [String : Any] {
+        } else if value is [String: Any] {
             return "a dictionary"
         } else {
             return "\(type(of: value))"
         }
     }
 }
-
-
--- a/Sources/XMLCoder/Decoder/XMLDecoder.swift
+++ b/Sources/XMLCoder/Decoder/XMLDecoder.swift
@@ -196,7 +196,19 @@ open class XMLDecoder {
     
     /// Contextual user-provided information for use during decoding.
     open var userInfo: [CodingUserInfoKey : Any] = [:]
-    
+
+	/// When an XML Element has both attributes and child elements, the CharData within the element will be keyed with the `characterDataToken`
+	/// - Note The following XML has both attribute data, child elements, and character data:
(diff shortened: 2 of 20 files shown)


Sources/XMLCoder/Decoder/DecodingErrorExtension.swift
```diff
@@ -23,7 +23,7 @@ internal extension DecodingError {
         let description = "Expected to decode \(expectation) but found \(_typeDescription(of: reality)) instead."
         return .typeMismatch(expectation, Context(codingPath: path, debugDescription: description))
     }
-
+
```
MaxDesiatov: This looks like trailing whitespaces on a blank line, was this committed by mistake or added by SwiftFormat? In either case it would be great to avoid these

Sources/XMLCoder/Decoder/XMLDecoder.swift
```diff
@@ -151,12 +152,12 @@ open class XMLDecoder {
                 stringKey.formIndex(before: &lastNonUnderscore)
             }
             
-            let keyRange = firstNonUnderscore...lastNonUnderscore
-            let leadingUnderscoreRange = stringKey.startIndex..<firstNonUnderscore
-            let trailingUnderscoreRange = stringKey.index(after: lastNonUnderscore)..<stringKey.endIndex
+            let keyRange = firstNonUnderscore ... lastNonUnderscore
+            let leadingUnderscoreRange = stringKey.startIndex ..< firstNonUnderscore
+            let trailingUnderscoreRange = stringKey.index(after: lastNonUnderscore) ..< stringKey.endIndex
```
MaxDesiatov: Adding spaces around range operators is an interesting style I haven't seen before, including this library. In general I'd prefer to stick to having no spaces around those, which is the same as what the Swift standard library does for example. Does SwiftFormat support that?

Tests/XMLCoderTests/CharacterDataDecodingTestCase.swift
```diff
@@ -29,4 +29,42 @@ class CharacterDataDecodingTestCase: DecodingTestCase {
...
+			<item ID="1542637462" Creator="Francisco Moya">
+			<attribute name="Xpos">0</attribute>
+			</item>
+			"""
+
+		// when
+		let result: Item = try! decode(xml)
+
+		// then
+		let expected = Item(id: "1542637462", creator: "Francisco Moya", children: nil)
```
MaxDesiatov: I'm a bit confused why `children` gets a `nil` value here, isn't it supposed to be `Attribute(name: "Xpos", value: "0")`?

qmoya: Hey Max! Thanks—by mistake, I pushed some things to this PR that should have gone elsewhere. My last intended commit was ‘Complete if #availables’. I’ll revert as appropriate and fix the automated formatting. Thanks!

Issue #17, opened 2018-12-15
XmlEncoder: ordering of elements


Ma-He:
Hi,
is there any possibility to set the order of attributes in an element while encoding a model-instance?
I'm trying to build a GPX reader/writer with this little awesome framework and have noticed that the order of my attributes in an element varies from encode to encode. Am I missing something?

Thanks in advance

MaxDesiatov:
Hi @Ma-He, thank you reporting this issue. There's a [`.sortedKeys`](https://github.com/MaxDesiatov/XMLCoder/blob/master/Sources/XMLCoder/Encoder/XMLEncoder.swift#L34) option that can be set for [`XMLEncoder.outputFormatting`](https://github.com/MaxDesiatov/XMLCoder/blob/master/Sources/XMLCoder/Encoder/XMLEncoder.swift#L202) property.

We plan to publish auto-generated docs for `XMLCoder` soon, which would make this option more visible to users.

Also, if you're using this option, you might need to switch to `master` branch as it was recently reported that this option was broken. This was fixed in #11 and #14 and merged to `master`.

regexident:
Sorting is not was OP is looking for. Ordering is. That is `<foo>…</foo>` has to appear before `<bar>…</bar>`, if `.foo` was encoded before `.bar`.

This is not supported currently as XMLCoder uses unordered `NSDictionary` in its internal keyed encoding container.

MaxDesiatov:
Thanks @regexident, I understand now. Still, my understanding was that the main concern is about determinism. I hope that `.sortedKeys` is able to solve that. I guess we have to transition to a tuple array for internal storage to maintain order as well.

regexident:
I know for sure as writing a `Codable`-powered GPX implementation was my reason for dabbling with XMLCoder (or rather it's predecessor XMLParser) as well.

GPX (like many other XML schemas) expects certain keys to have a specific order:

Search for `<!--  elements must appear in this order  -->` in the [GPX 1.1 Schema](https://www.topografix.com/GPX/1/1/gpx.xsd).

regexident:
Nitpick: It's not the order attributes that's unsupported, it's the order of elements.

The XML specification explicitly defines the order of attributes to be of no importance:

[...]

PR #19, opened 2018-12-18
Overhaul internal representation, replacing `NS…` with `…Box` types

regexident:
This PR removes any remaining use of `NS…Array`/`NS…Dictionary`/`NSNumber`/`NSDecimalNumber`/`NSNull`/… (as storage) from the code-base.

It introduces an internal type …

```swift
internal enum Box {
    case null(NullBox)
    case bool(BoolBox)
    case decimal(DecimalBox)
    case signedInteger(SignedIntegerBox)
    case unsignedInteger(UnsignedIntegerBox)
    case floatingPoint(FloatingPointBox)
    case string(StringBox)
    case array(ArrayBox)
    case dictionary(DictionaryBox)
}
```

… as well as accompanying variant box types, replacing use of `Any`/`NS…`.

👷🏻‍♀️It improves type-safety by reducing use of `Any` (from 60 down to 19 occurrences) as well as `NSObject` (from 37 down to 1 occurrence).

🏗It further more levels the field for improvements/additions such as [support for order-preserving elements & attributes](https://github.com/MaxDesiatov/XMLCoder/issues/17).

💡Thanks to encapsulation we can now safely change the inner logic of `DictionaryBox` to retain insertion order.

**Edit**:

We ended up replacing aforementioned `enum Box` with a protocol:

```swift
protocol Box {
    var isNull: Bool { get }
    var isFragment: Bool { get }
    
    func xmlString() -> String?
}
```


Sources/XMLCoder/Auxiliaries/XMLElement.swift
```diff
@@ -0,0 +1,280 @@
...
+internal class _XMLElement {
+    static let attributesKey = "___ATTRIBUTES"
+    static let escapedCharacterSet = [("&", "&amp"), ("<", "&lt;"), (">", "&gt;"), ("'", "&apos;"), ("\"", "&quot;")]
+    
+    var key: String
+    var value: String?
+    var attributes: [String: String] = [:]
+    var children: [String: [_XMLElement]] = [:]
+    
+    internal init(key: String, value: String? = nil, attributes: [String: String] = [:], children: [String: [_XMLElement]] = [:]) {
```
MaxDesiatov: I think this could lose `internal`, which is the default anyway.

Sources/XMLCoder/Auxiliaries/XMLElement.swift
```diff
@@ -0,0 +1,280 @@
...
+        self.key = key
+        self.value = value
+        self.attributes = attributes
+        self.children = children
+    }
+    
+    static func createRootElement(rootKey: String, object: ArrayBox) -> _XMLElement? {
```
MaxDesiatov: These static functions return the same type they're declared on. This is a good chance to stick to the convention and rewrite this as an optional initializer. This is applicable to all other `createRootElement` functions, which I see delegate to `modifyElement` though. Maybe change to initializers should happen in a separate PR when `_XMLElement` becomes a `struct`?

Sources/XMLCoder/Auxiliaries/XMLElement.swift
```diff
@@ -0,0 +1,280 @@
...
+        }
+        
+        if let parentElement = parentElement, let key = key {
+            parentElement.children[key] = (parentElement.children[key] ?? []) + [element]
+        }
+    }
+    
+    fileprivate static func createElement(parentElement: _XMLElement, key: String, object: NullBox) {
+        let element = _XMLElement(key: key)
+        parentElement.children[key] = (parentElement.children[key] ?? []) + [element]
```
MaxDesiatov: couldn't this be simplified as `parentElement.children[key]?.append(element)`?

Sources/XMLCoder/Auxiliaries/XMLElement.swift
```diff
@@ -0,0 +1,280 @@
...
+    }
+    
+    fileprivate static func modifyElement(element: _XMLElement, parentElement: _XMLElement?, key: String?, object: DictionaryBox) {
+        let attributesBox = object[_XMLElement.attributesKey]?.dictionary
+        element.attributes = attributesBox?.unbox().mapValues { $0.xmlString } ?? [:]
+        
+        let objects = object.filter { key, _value in key != _XMLElement.attributesKey }
+        
+        for (key, box) in objects {
+            switch box {
```
MaxDesiatov: I'm not entirely convinced that this `switch` and duplication of `static func createElement(parentElement: _XMLElement, key: String, object:)` for every box type is the best solution. I'd rather prefer something like this:

```swift
protocol XMLBox {
  var xmlString: String { get }
}

// Every box type conforms to `Box` and implements `xmlString`

// ...
            static func createElement(parentElement: _XMLElement, key: String, object: Box) {
	        let element = _XMLElement(key: key, value: object.xmlString)
	        parentElement.children[key] = (parentElement.children[key] ?? []) + [element]
	    }
//...
```

This `switch` would go away then as well.

Sources/XMLCoder/Auxiliaries/XMLElement.swift
```diff
@@ -0,0 +1,280 @@
...
+    fileprivate static func createElement(parentElement: _XMLElement, key: String, object: StringBox) {
+        let element = _XMLElement(key: key, value: object.xmlString)
+        parentElement.children[key] = (parentElement.children[key] ?? []) + [element]
+    }
+    
+    fileprivate static func createElement(parentElement: _XMLElement, key: String, object: ArrayBox) {
+        let objects = object.unbox()
+        
+        for box in objects {
+            switch box {
```
MaxDesiatov: Same comment as above, a `protocol` would work better than a `switch` in my opinion.

Sources/XMLCoder/Box/ArrayBox.swift
```diff
@@ -0,0 +1,81 @@
...
+//  ArrayBox.swift
+//  XMLCoderPackageDescription
+//
+//  Created by Vincent Esche on 11/20/18.
+//
+
+import Foundation
+
+// Minimalist implementation of an order-preserving unkeyed box:
+internal class ArrayBox {
```
MaxDesiatov: Is there a specific reason for this to be a `class` instead of `struct`? Also I think this could conform to `Collection` or `Sequence` and get `map` and `filter` implementations for free from protocol extensions already available in the Swift standard library.

Sources/XMLCoder/Box/Box.swift
```diff
@@ -0,0 +1,238 @@
...
+    
+    var dictionary: DictionaryBox? {
+        guard case .dictionary(let box) = self else {
+            return nil
+        }
+        return box
+    }
+    
+    var xmlString: String {
+        switch self {
```
MaxDesiatov: I think this `switch` could be avoided with a `Box` protocol as described in comments above.

Sources/XMLCoder/Box/Box.swift
```diff
@@ -0,0 +1,238 @@
...
+    
+    var isNull: Bool {
+        guard case .null(_) = self else {
+            return false
+        }
+        return true
+    }
+    
+    var isFragment: Bool {
+        switch self {
```
MaxDesiatov: `isFragment` could probably also be a property on a new `Box` protocol that would help to avoid this `switch`

Sources/XMLCoder/Box/Box.swift
```diff
@@ -0,0 +1,238 @@
...
+    
+    init(_ unboxed: [Box]) {
+        self = .array(.init(unboxed))
+    }
+    
+    init(_ unboxed: [String: Box]) {
+        self = .dictionary(.init(unboxed))
+    }
+    
+    internal func unbox() throws -> Bool? {
```
MaxDesiatov: probably could remove `internal` in this function and below?

Sources/XMLCoder/Box/Box.swift
```diff
@@ -0,0 +1,238 @@
...
+    internal func unbox() throws -> [String: Box]? {
+        switch self {
+        case .dictionary(let box): return box.unbox()
+        case _: return nil
+        }
+    }
+}
+
+extension Box: Equatable {
+    static func == (lhs: Box, rhs: Box) -> Bool {
```
MaxDesiatov: hm, wouldn't `Equatable` implementation for `Box` derive automatically with the correct implementation? My thinking was that you can get a compiler-generated `Equatable` for all enums with associated values that conform to `Equatable`, which I think those associated values should have anyway

Sources/XMLCoder/Box/DictionaryBox.swift
```diff
@@ -0,0 +1,89 @@
...
+//  DictionaryBox.swift
+//  XMLCoderPackageDescription
+//
+//  Created by Vincent Esche on 11/19/18.
+//
+
+import Foundation
+
+// Minimalist implementation of an order-preserving keyed box:
+internal class DictionaryBox {
```
MaxDesiatov: same comment on `struct` vs `class` and lack of conformance to Swift standard library protocols as for `ArrayBox`

Sources/XMLCoder/Auxiliaries/XMLElement.swift
regexident: The whole current `createRootElement`/`modifyElement` shenanigans is rather cryptic, imho.

As such I'm all for cleaning up `_XMLElement`.

Sources/XMLCoder/Auxiliaries/XMLElement.swift
regexident: Good point. I like to put `internal` to make the intention explicit (vs. "did I forget to make this `public/…`? 🤔"). But I forgot to make things consistent. I'm happy with removing it though.

Sources/XMLCoder/Auxiliaries/XMLElement.swift
MaxDesiatov: cool, I think `_XMLElement` with these static functions could stay as they are within this PR and change later in a separate one to make the review easier

Sources/XMLCoder/Auxiliaries/XMLElement.swift
regexident: I think so, yes.

Sources/XMLCoder/Auxiliaries/XMLElement.swift
```diff
@@ -0,0 +1,280 @@
+//
+//  XMLElement.swift
+//  XMLCoder
+//
+//  Created by Vincent Esche on 12/18/18.
+//
+
+import Foundation
+
+internal class _XMLElement {
```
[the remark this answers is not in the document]
regexident: My guess is that some of the logic expects reference semantics. The `func modifyElement` in particular. One would have to pass it as `inout` if made into a struct, I think.

Sources/XMLCoder/Box/Box.swift
regexident: Fixed.

MaxDesiatov: Having it all in my head at once, I think if you require `Equatable` for all `Box` implementors you could avoid this custom implementation too. Like this:

```
protocol Box: Equatable {
   var xmlString: String { get }
   var isFragment: Bool { get }
}

extension Array: Box {
//...
}

extension Bool: Box {
//...
}
```

If you use standard `Array` instead of new `ArrayBox`, you'd have `Equatable` implementations on these wrapped types already.

Sources/XMLCoder/Box/DictionaryBox.swift
MaxDesiatov: Again, re-reviewing this, I think we could just get away with `extension Dictionary: Box { ... }` if `Box` was a protocol.

Sources/XMLCoder/Box/Box.swift
regexident: Fixed.

Sources/XMLCoder/Auxiliaries/XMLElement.swift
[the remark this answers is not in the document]
regexident: Confirmed: `_XMLKeyedDecodingContainer` and `_XMLUnkeyedDecodingContainer` hold a reference to their internal container:

```swift
/// A reference to the container we're reading from.
private let container: ArrayBox
```

```swift
/// A reference to the container we're reading from.
private let container: DictionaryBox
```

The `_XMLReferencingEncoder` also seems to rely on reference semantics.

Sources/XMLCoder/Auxiliaries/XMLElement.swift
regexident: Fixed.

Sources/XMLCoder/Auxiliaries/XMLElement.swift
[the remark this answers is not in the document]
MaxDesiatov: oh, I'm not even entirely sure if we necessarily need `_XMLReferencingEncoder` in the end, which is only used in `superEncoder`. I think if an encoder returned from `superEncoder` operates on a copy that would make things even cleaner? And if we're concerned with a performance hit caused by copies we could consider other optimisations or using something like `__consuming` from Swift 5.0. My current understanding is that `_XMLReferencingEncoder` relies on modifying containers referenced by other encoders, which I hope could be avoided and make the code and architecture cleaner. Looking into it a bit more and would be very happy to hear your thoughts on `_XMLReferencingEncoder`.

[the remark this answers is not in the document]
MaxDesiatov: From the Apple's docs on `superEncoder`:

> Stores a **new** nested container for the given key and returns a **new** encoder instance for encoding super into that container.

Which I understand as there's no expectation from the standard library for that encoder to hold any references.

Looks like `_XMLReferencingEncoder` has no test coverage, I'll add a test with a codable class hierarchy and experiment with removing `_XMLReferencingEncoder` and just creating a copy of the container in a separate PR.

regexident:
The problems I see with turning `Box` into a protocol are these:

The protocol(s) would have to look something like this to work:

```
protocol AnyBoxProtocol: Equatable {
    var xmlString: String { get }
}

protocol BoxProtocol: AnyBoxProtocol {
    associatedtype Unboxed
    
    init(_ unboxed: Unboxed)
    
    func unbox() -> Unboxed
}
```

But in the case of `SignedIntegerBox`/`UnsignedIntegerBox`/`FloatingPointBox` the `BoxProtocol` couldn't even be implemented properly as their `init(_:)` and `func unbox` differ significantly due to covering whole type families, not just a single wrapped type.

If Swift allowed for multiple protocol conformances we'd be fine. Alas it doesn't.

Another benefit of using a enum is that we get a guarantee for exhaustive `switch`es.
Adding a new box type some time in the future is far less brittle that way, than it is with type erasure (be it `Any` or `protocol Box`).

Sources/XMLCoder/Box/DictionaryBox.swift
regexident: I chose to keep the API surface area small by only implementing functions that are actually used. There is imho no point in covering all of `Array`/`Dictionary`'s API.

Sources/XMLCoder/Box/Box.swift
regexident: > If you use standard Array instead of new ArrayBox, you'd have Equatable implementations on these wrapped types already.

One of my main driving reasons for using Box types is to encapsulate the behavior. I want to be able to change the internal behavior without having to edit dozens of lines of code sprinkled all over the project.

Case in point: Changing `DictionaryBox` to have `OrderedDictionary` semantic, next. Which is my main motivation for this PR.

---

I'm also think that the whole logic within all those `func box`/`func unbox` in `_XMLEncoder`/`_XMLDecoder` should be moved into their respective individual Box types to make things easier to test and more organized/encapsulated.

Right now `XMLEncoder.swift`/`XMLDecoder.swift` are 600/800loc monsters (even after I moved a whole bunch of code out there already!) that will be a pain to maintain in the future.

regexident:
Also could Apple please get rid of the plague that is `. xcodeproj` already?
Those merge conflicts are the worst. It's almost 2019, damnit! 😩

MaxDesiatov:
Great point, I think we should consider moving all custom settings to `Package.xcconfig` like [I've done in `CoreXLSX`](https://github.com/MaxDesiatov/CoreXLSX/blob/master/Package.xcconfig) and get rid of `.xcodeproj` altogether in the repository encouraging users to run `swift package generate-xcodeproj --xcconfig-overrides Package.xcconfig` on their side when needed. The only problem is baselines which are stored within `.xcodeproj` and I'm not sure how useful benchmarks are without baselines... 🤔

Sources/XMLCoder/Auxiliaries/XMLElement.swift
regexident: Fixed.

regexident:
Commit 7fe7984 just removed `enum Box` in favor of `protocol Box`. 🙂

Sources/XMLCoder/Box/ArrayBox.swift
```diff
@@ -0,0 +1,102 @@
...
+    }
+    
+    var isFragment: Bool {
+        return false
+    }
+    
+    var xmlString: String? {
+        return nil
+
+//        let strings: [String] = self.unboxed.map { box in
```
MaxDesiatov: I assume this is the correct implementation that should be uncommented?

Sources/XMLCoder/Box/DictionaryBox.swift
```diff
@@ -0,0 +1,105 @@
...
+    }
+    
+    var isFragment: Bool {
+        return false
+    }
+    
+    var xmlString: String? {
+        return nil
+        
+//        let elements = self.unboxed.sorted { $0.0 < $1.0 }
```
MaxDesiatov: Would be great if this is uncommented or removed before merge

Sources/XMLCoder/Box/UnsignedIntegerBox.swift
```diff
@@ -0,0 +1,42 @@
...
+
+import Foundation
+
+internal struct UnsignedIntegerBox: Equatable {
+    typealias Unboxed = Int
+    
+    let unboxed: Unboxed
+    
+    init<Integer: UnsignedInteger>(_ unboxed: Integer) {
+        self.unboxed = Unboxed(unboxed)
```
MaxDesiatov: hm, does this mean that we'll truncate `Int64` on platforms where `Int` is actually `Int64`, like Apple Watch?

Sources/XMLCoder/Auxiliaries/XMLHeader.swift
```diff
@@ -0,0 +1,47 @@
...
+    /// indicates whether a document relies on information from an external source.
+    public let standalone: String?
+    
+    public init(version: Double? = nil, encoding: String? = nil, standalone: String? = nil) {
+        self.version = version
+        self.encoding = encoding
+        self.standalone = standalone
+    }
+    
+    internal func isEmpty() -> Bool {
```
MaxDesiatov: another use of `internal` here

Sources/XMLCoder/Auxiliaries/XMLHeader.swift
```diff
@@ -0,0 +1,47 @@
...
+        return version == nil && encoding == nil && standalone == nil
+    }
+    
+    internal func toXML() -> String? {
```
MaxDesiatov: use of `internal`

Sources/XMLCoder/Auxiliaries/XMLStackParser.swift
```diff
@@ -0,0 +1,98 @@
...
+//  XMLStackParser.swift
+//  XMLCoder
+//
+//  Created by Shawn Moore on 11/14/17.
+//  Copyright © 2017 Shawn Moore. All rights reserved.
+//
+
+import Foundation
+
+internal class _XMLStackParser: NSObject, XMLParserDelegate {
```
MaxDesiatov: `internal`

Sources/XMLCoder/Box/BoolBox.swift
```diff
@@ -0,0 +1,42 @@
+//
+//  BoolBox.swift
+//  XMLCoder
+//
+//  Created by Vincent Esche on 12/17/18.
+//
+
+import Foundation
+
+internal struct BoolBox: Equatable {
```
MaxDesiatov: use of `internal`

Sources/XMLCoder/Auxiliaries/String+Extensions.swift
```diff
@@ -0,0 +1,20 @@
...
+//  String+Extensions.swift
+//  XMLCoder
+//
+//  Created by Vincent Esche on 12/18/18.
+//
+
+import Foundation
+
+extension String {
+    internal func escape(_ characterSet: [(character: String, escapedCharacter: String)]) -> String {
```
MaxDesiatov: `internal` can be removed for consistency

Sources/XMLCoder/Box/ArrayBox.swift
```diff
@@ -0,0 +1,102 @@
...
+//                }
+//            case _:
+//                return box.xmlString
+//            }
+//        }
+//        return "[" + strings.joined(separator: ", ") + "]"
+    }
+}
+
+//extension ArrayBox: Equatable {
```
MaxDesiatov: code commented out

Sources/XMLCoder/Box/DictionaryBox.swift
```diff
@@ -0,0 +1,105 @@
...
+//                } else {
+//                    return nil
+//                }
+//            }
+//        }
+//        return "[" + strings.joined(separator: ", ") + "]"
+    }
+}
+
+//extension DictionaryBox: Equatable {
```
MaxDesiatov: commented extension? Which can probably be derived automatically by the compiler if needed

Sources/XMLCoder/Auxiliaries/XMLHeader.swift
regexident: There are 94 uses of `internal`. I removed all that were added by this PR.

MaxDesiatov: awesome, that's great, thank you!

Sources/XMLCoder/Box/DictionaryBox.swift
regexident: It can't any more, due to `Box` not being `Box: Equatable`. The latter would turn it into a PAT, rendering it unusable for our purposes.

But I only needed the `Equatable` for testing anyway. Using a workaround there now.

Sources/XMLCoder/Box/ArrayBox.swift
regexident: I only needed the `Equatable` for testing. Using a workaround there now.

Sources/XMLCoder/Box/DictionaryBox.swift
MaxDesiatov: that makes sense, thank you. Would be great if the workaround in the test has a comment explaining that it exists only because of the PAT issue so that in the future we don't forget why it was done that way.

Sources/XMLCoder/Auxiliaries/XMLHeader.swift
regexident: Wait, what I meant by that is "there still are 94 occurrences of `internal` after removing (what should be) all I added". 😅

Removing 94 occurrences of `internal` sprinkled all over `XMLCoder` adds a lot of noise to this already rather large PR.
I'd suggest doing a cleanup PR afterwards that does just that. What do you think?

Sources/XMLCoder/Box/DictionaryBox.swift
regexident: I'm basically doing this:

```swift
func testUnbox() {
    let unboxed = box.unbox()
    XCTAssertEqual(unboxed.count, 2)
    XCTAssertEqual(unboxed[0] as? StringBox, StringBox("foo"))
    XCTAssertEqual(unboxed[1] as? IntBox, IntBox(42))
}
```

rather than this:

```swift
func testUnbox() {
    let unboxed = box.unbox()
    XCTAssertEqual(unboxed, [StringBox("foo"), IntBox(42)])
}
```

No magic in there. Just avoiding the type-erasure of `Box`.

Sources/XMLCoder/Box/ArrayBox.swift
regexident: Yes, this is the expected behavior. We're not calling `.xmlString` on arrays or dictionaries, but rather delegating to their elements.

Commented out code has been removed.

Sources/XMLCoder/Box/DictionaryBox.swift
regexident: Yes, this is the expected behavior. We're not calling `.xmlString` on arrays or dictionaries, but rather delegating to their elements.

Commented out code has been removed.

Sources/XMLCoder/Box/UnsignedIntegerBox.swift
regexident: I changed it to using `Int64` and `UInt64` instead of `Int` and `UInt` as 
`Unbounded`. This should work with AppleWatch, right?

Fixed.

Sources/XMLCoder/Auxiliaries/XMLHeader.swift
MaxDesiatov: all good, thanks

Tests/XMLCoderTests/Box/DataBoxTests.swift
```diff
@@ -0,0 +1,72 @@
...
+        ]
+        
+        for (string, fromXMLString) in values {
+            let box = fromXMLString(string)!
+            XCTAssertEqual(box.xmlString(), string)
+        }
+    }
+    
+    func testValidValues() {
+        typealias Constructor = (String) -> Boxed?
```
MaxDesiatov: this typealias looks unused

Tests/XMLCoderTests/Box/DataBoxTests.swift
```diff
@@ -0,0 +1,72 @@
...
+        ]
+        
+        for (string, fromXMLString) in values {
+            let box = fromXMLString(string)
+            XCTAssertNotNil(box)
+        }
+    }
+    
+    func testInvalidValues() {
+        typealias Constructor = (String) -> Boxed?
```
MaxDesiatov: this typealias is used once, but it's equivalent to `FromXMLString`?

Tests/XMLCoderTests/Box/DataBoxTests.swift
```diff
@@ -0,0 +1,72 @@
...
+        let values: [(String, FromXMLString)] = [
+            ("bG9yZW0gaXBzdW0=", base64),
+            ("ZG9sb3Igc2l0IGFtZXQ=", base64),
+            ("Y29uc2VjdGV0dXIgYWRpcGlzY2luZyBlbGl0", base64),
+        ]
+        
+        for (string, fromXMLString) in values {
+            let box = fromXMLString(string)!
+            XCTAssertEqual(box.xmlString(), string)
+        }
```
MaxDesiatov: sorry, I find this whole part with `FromXMLString` a bit hard to follow. Could this test and others in this test case be simplified like this?

```swift
let testStrings = [ 
  "bG9yZW0gaXBzdW0=",  
  "ZG9sb3Igc2l0IGFtZXQ=",
  "Y29uc2VjdGV0dXIgYWRpcGlzY2luZyBlbGl0",
]

for string in testStrings {
  XCTAssertEqual(Boxed(base64: string).xmlString(), string)
}
```

I understand that it tries to unify tests with tests for other box types, but maybe then it should have a comment clarifying the testing approach and making the typealias visible to all those tests or to make the test much simpler.

regexident:
Rebased onto `master`.

regexident:
Phew, this PR ended up quite bigger than anticipated, but it's worth it, I think.

The separation and encapsulation, as well as addition of finely-grained unit tests should give us a good and solid foundation for future improvements and bug fixes.

Thanks for your comments, @MaxDesiatov!

PR #20, opened 2018-12-18
Added benchmark to RJI test suite

regexident:


MaxDesiatov:
Thank you @regexident, I'll update the sample XML to the new version in `master`, which is no longer stored in `.xml` file. I had to change that in #18 as generation of Jazzy docs fails because of cruft accumulated in `XMLCoder.xcodeproj`. We have to regenerate the project file with `swift package generate-xcodeproj`. This auto-generation misses files that don't have `.swift` extensions unfortunately, which means we have missing sample XML file in the test bundle. The sample XML was moved to the `.swift` file from `.xml` file.

PR #21, opened 2018-12-18
Fix RJI test encoding

### LINKED ISSUE #15: Investigate XML decoding in RJI test case
Currently RJI RSS test case is able to decode the cleaned up XML successfully (multiple "author" elements were removed to be investigated later). The main problem is that when RJI value is encoded it can't be decoded back again without errors, there were attempts to add this to the test with [this commented code](https://github.com/MaxDesiatov/XMLCoder/blob/7e13f4bd5407b83408bc6fc86b9aa31a155b53ce/Tests/XMLCoderTests/RJITest.swift#L157).

hodovani:
Resolve #15

### DIFF
--- a/Sources/XMLCoder/Auxiliaries/XMLElement.swift
+++ b/Sources/XMLCoder/Auxiliaries/XMLElement.swift
@@ -9,7 +9,7 @@ import Foundation
 
 internal class _XMLElement {
     static let attributesKey = "___ATTRIBUTES"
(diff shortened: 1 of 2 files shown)


MaxDesiatov:
@hodovani to make this a bit easier to debug, I'd suggest creating a new variable with a single item or two from `rjiSampleXML` and test if those encode correctly. We could also eliminate xml elements/attributes and struct properties one by one to understand what exactly is failing.

PR #26, opened 2018-12-20
Clean up `XMLStackParser`

regexident:
This PR consists of mostly migrations from `if let` to `guard let` as well as other semantics-preserving changed to keep the conditional nesting low and make the control flow clear (i.e. the happy/expected path is the one with the least detours into `if` or `if let` blocks).

### DIFF
--- a/Sources/XMLCoder/Auxiliaries/XMLElement.swift
+++ b/Sources/XMLCoder/Auxiliaries/XMLElement.swift
@@ -209,3 +209,17 @@ internal class _XMLElement {
         return string
     }
 }
+
+extension _XMLElement: Equatable {
+    static func == (lhs: _XMLElement, rhs: _XMLElement) -> Bool {
+        guard
+            lhs.key == rhs.key,
+            lhs.value == rhs.value,
+            lhs.attributes == rhs.attributes,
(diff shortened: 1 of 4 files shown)


Sources/XMLCoder/Auxiliaries/XMLElement.swift
```diff
@@ -209,3 +209,21 @@ internal class _XMLElement {
...
+        guard lhs.key == rhs.key else {
+            return false
+        }
+        guard lhs.value == rhs.value else {
+            return false
+        }
+        guard lhs.attributes == rhs.attributes else {
+            return false
+        }
+        guard lhs.children == rhs.children else {
```
MaxDesiatov: would it be possible to merge all of the equality comparisons into a single `guard` block given that all of them do the same `return false`?

regexident: Sure. Fixed.

MaxDesiatov:
@hodovani do you have any comments or questions about he PR? this can be merged otherwise.

hodovani:
No question from me, thank you @regexident!

PR #31, opened 2018-12-21
Fixed a bug that would cause `XMLEncoder` to throw on `Encodable`s encoding nothing

regexident:
By encoding via `func encode(_:withRootKey:header:)` we always get an XML root node with label `rootKey`, so not encoding anything within `func encode(to:)` is fine in XML (unlike JSON).


MaxDesiatov:
@regexident we got an intermittent problem with force-pushing to the same branch, but I think I got it fixed now, sorry about that.

PR #33, opened 2018-12-22
Let compiler synthesize Equatable conformance for _XMLElement

jsbean:
As `_XMLElement` was made a `struct` in 1f0cb9d91178678b64e251a5b761dbee7f519445, we can let the compiler synthesize `Equatable` conformance for free.

This PR removes the custom `_XMLElement: Equatable` conformance implementation.


jsbean:
Ah, so SwiftLint does not like the empty `: Equatable { }` notation. Otherwise, the tests pass.

Would you prefer the conformance to be placed up top (`struct _XMLElement: Equatable {`) or in some other fashion?

PR #34, opened 2018-12-22
Add dedicated benchmarking test suite

regexident:
Admittedly the `_ closure: (() -> ()) -> ()` + `closure { … }` mumbo jumbo isn't the prettiest code I've ever written, but then again XCTest isn't a particularly usable testing framework either. 🤷🏻‍♂️

Xcode attaches the measurements popover to literal code occurrences of `self.measure { … }` itself, making the whole mess required in the first place. 🙄
(All while only supporting a single `self.measure { … }` per unit test. 🤦🏻‍♂️)

### DIFF
--- a/Tests/XMLCoderTests/BenchmarkTests.swift
+++ b/Tests/XMLCoderTests/BenchmarkTests.swift
@@ -28,11 +28,11 @@ class BenchmarkTests: XCTestCase {
         return try encoder.encode(container, withRootKey: "container")
     }
 
-    func testEncoding<T: Codable>(with value: T, _ closure: (() -> Void) -> Void) throws {
+    func testEncoding<T: Codable>(with value: T, _ closure: (() -> ()) -> ()) throws {
         try testEncoding(of: T.self, with: value, closure)
     }
 
-    func testEncoding<T: Codable>(of _: T.Type, with value: T, _ closure: (() -> Void) -> Void) throws {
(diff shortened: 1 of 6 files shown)


Tests/XMLCoderTests/BenchmarkTests.swift
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    func testDecoding<T: Codable>(with value: T, _ closure: (() -> Void) -> Void) throws {
+        try testDecoding(of: T.self, with: value, closure)
+    }
+    
+    func testDecoding<T: Codable>(of type: T.Type, with value: T, _ closure: (() -> Void) -> Void) throws {
+        let data: Data = try encodedContainer(of: T.self, with: value)
+        let string = String(data: data, encoding: .utf8)
+        print(string!)
```
MaxDesiatov: It would be great if we could avoid using `print` in testing code.

Tests/XMLCoderTests/BenchmarkTests.swift
```diff
@@ -0,0 +1,225 @@
...
+
```
MaxDesiatov: there is a trailing whitespace on this line and a few other lines

MaxDesiatov:
hm, now a benchmark test fails on CI...

regexident:
Yeah, I noticed that, too. 🤨
It didn't before, but that's because I had a bug in the tests only ever coding `null`s.

Issue #25, opened 2018-12-18
Array of enums with associated values


jsbean:
Thanks for taking up the maintenance on this project!

I am working with an XML schema which requires support for heterogeneous arrays of values within a set of specified types.

Consider the following enum which defines each of the types allowable:

```Swift
enum AB {
    struct A: Decodable { let value: Int }
    struct B: Decodable { let value: String }
    case a(A)
    case b(B)   
}
```

Currently, the `Decodable` implementation is as follows (based on [objc.io/codable-enums](https://www.objc.io/blog/2018/01/23/codable-enums/)):

```Swift
extension AB: Decodable {
    enum CodingKeys: String, CodingKey { case a, b }
    init(from decoder: Decoder) throws {
        let container = try decoder.container(keyedBy: CodingKeys.self)
        do {
            self = .a(try container.decode(A.self, forKey: .a))
        } catch {
            self = .b(try container.decode(B.self, forKey: .b))
        }
    }
}
```

A minimal example of the source XML could look something like this (though there could be arbitrary amounts of either type, in any order):

```Swift
let xml = """
<?xml version="1.0" encoding="UTF-8"?>
<container>
    <a value="42"></a>
    <b value="forty-two"></b>
</container>
"""
```

Upon decoding it with the `XMLDecoder`:

```Swift
let decoder = XMLDecoder()
let abs: [AB] = try! decoder.decode([AB].self, from: xml.data(using: .utf8)!)
```

We get the following:

```Swift
// => [.a(AB.A(value: 42))]
```

The desired result should be:

```Swift
// => [.a(AB.A(value: 42)), .b(AB.B(value: "forty-two"))]
```

I have tried many combinations of keyed, unkeyed, singleValue, and nested containers, yet haven't found a solution that treats the data as an `Array` rather than a `Dictionary`.

Let me know if I am holding this wrong (at either the `Decodable` or the `XMLDecoder` stages), or if this could be a current limitation of the `XMLCoder` implementation. I'd be happy to contribute, but if anyone notices something glaring here, this would be very helpful.

MaxDesiatov:
Hi @jsbean, first of all thanks a lot for the report and sorry for the delay caused by our current focus on test coverage in `XMLCoder`. We've finally reached ~75% coverage and are ready to look into more bugfixes and disruptive changes in the codebase. 

[...]

Issue #12, opened 2018-12-02
Can’t reach an XML value


qmoya:
Hi there! I can’t find the right data structure to decode the following XML string:

    <something name="Some name">Some value</something>

I ran out of ideas. May anyone give me a hint?

Thanks!

MaxDesiatov:
This is now supported in `master` with #73 merged. As usual please feel free to reopen this issue if you think there's anything missing.

thecb4:
Trying to use master with the following and still getting an error
 
<preview_image_time format="24/999 1000/nonDrop">00:00:17:01</preview_image_time>

------ 

struct PreviewImageTime: Codable, DynamicNodeEncoding {
  var format: String
  var value: String

  enum CodingKeys: String, CodingKey {
    case format
    case value = ""
  }

  static func nodeEncoding(forKey key: CodingKey) -> XMLEncoder.NodeEncoding {
    switch key {
      case CodingKeys.format:
        return .attribute
      default:
        return .element
    }
  }
}
------------------------

could not parse
typeMismatch(Swift.Dictionary<Swift.String, Any>, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "software", intValue: nil), CodingKeys(stringValue: "software_metadata", intValue: nil), CodingKeys(stringValue: "versions", intValue: nil), CodingKeys(stringValue: "version", intValue: nil), XMLKey(stringValue: "0", intValue: 0), CodingKeys(stringValue: "locales", intValue: nil), CodingKeys(stringValue: "locale", intValue: nil), XMLKey(stringValue: "0", intValue: 0), CodingKeys(stringValue: "app_previews", intValue: nil), CodingKeys(stringValue: "app_preview", intValue: nil), XMLKey(stringValue: "0", intValue: 0), CodingKeys(stringValue: "preview_image_time", intValue: nil)], debugDescription: "Expected to decode Dictionary<String, Any> but found SharedBox<UnkeyedBox> instead.", underlyingError: nil))

JoeMatt:
@thecb4 What happens if you change your string values to something without colons or backslashes.

I think I ran into something similar with UUID strings that had colons. I believe the decoder is attempting to decode a dictionary before a string, based on the ability to convert "k : v " strings to dictionaries.

If that's a case, a workaround may be to provide your own decoder that specifies the type to try.

I would just try a simple test case first with simple strings to see if that is indeed the issue. I have some example code that might help if that is it.

[...]

PR #36, opened 2018-12-22
Keyed/unkeyed boxes as structs

regexident:
Not sure yet, if this will actually lead to an improvement of performance.

Running the benchmark from https://github.com/MaxDesiatov/XMLCoder/pull/34 should give a better image of the situation.

🏗Consider this an experimental WIP.

### DIFF
--- a/Sources/XMLCoder/Auxiliaries/XMLElement.swift
+++ b/Sources/XMLCoder/Auxiliaries/XMLElement.swift
@@ -66,6 +66,10 @@ struct _XMLElement {
 
     init(key: String, box: Box) {
         switch box {
+        case let sharedUnkeyedBox as SharedBox<UnkeyedBox>:
+            self.init(key: key, box: sharedUnkeyedBox.unbox())
+        case let sharedKeyedBox as SharedBox<KeyedBox>:
+            self.init(key: key, box: sharedKeyedBox.unbox())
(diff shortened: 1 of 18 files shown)


regexident:
Upon further inspection (ie. running https://github.com/MaxDesiatov/XMLCoder/pull/34 repeatedly) the performance seems to have remained stable.

I'd thus propose to merge this PR as it unifies the representation of `T` where `T: Box` and makes it explicit where shared reference semantics is expected.

regexident:
Any chance to have this merged, @MaxDesiatov/@hodovani?

MaxDesiatov:
@regexident I've fixed the conflicts and there's still a `superEncoder` test that fails here. But the most interesting part was that the performance is consistently 10-30% worse, could this be caused by copies of value types? Should we postpone this PR until Swift 5.0 is released or whatever the version is that allows us to improve performance of value types?

screen shot 2019-01-10 at 12 13 14

MaxDesiatov:
All tests are fixed now and since this is a dependency of #55, I'm inclined to merge both of these. After reviewing related issues I'd prefer XMLCoder to be slightly slower, but with #12, #17 and #25 fixed than the opposite 😞

regexident:
Thanks @MaxDesiatov!

PR #37, opened 2018-12-22
[RFC] Add OrderedDictionary

jsbean:
This PR adds an `OrderedDictionary` structure to the project, though does not use it for anything yet.

This should be useful for maintaining order of ~~attributes~~ elements (#17), and was useful in an experimental outing to solve #25, and I can imagine it being useful for #12, as well.

I thought I'd propose this here to start some conversation about these issues and how to approach solving them. If this won't be of help, feel free to close!

### DIFF
--- a/Sources/XMLCoder/Auxiliaries/OrderedDictionary.swift
+++ b/Sources/XMLCoder/Auxiliaries/OrderedDictionary.swift
@@ -102,7 +102,7 @@ struct OrderedDictionary <Key: Hashable, Value> {
         by areInIncreasingOrder: ((key: Key, value: Value), (key: Key, value: Value)) throws -> Bool
     ) rethrows -> OrderedDictionary
     {
-        return OrderedDictionary(uniqueKeysWithValues: try self.sorted(by: areInIncreasingOrder))
+        return OrderedDictionary(uniqueKeysWithValues: try sorted(by: areInIncreasingOrder))
     }
 }
 
--- a/Sources/XMLCoder/Auxiliaries/OrderedDictionary.swift
+++ b/Sources/XMLCoder/Auxiliaries/OrderedDictionary.swift
@@ -14,7 +14,7 @@ struct OrderedDictionary <Key: Hashable, Value> {
     var keys: [Key] = []
 
     /// - Returns: The values in order.
-    public var values: [Value] {
+    var values: [Value] {
         return keys.map { self[$0]! }
     }
 
@@ -63,7 +63,7 @@ struct OrderedDictionary <Key: Hashable, Value> {
     // MARK: - Subscripts
 
     /// - Returns: Value for the given `key`, if available. Otherwise, `nil`.
-    public subscript(key: Key) -> Value? {
+    subscript(key: Key) -> Value? {
         get {
             return unordered[key]
         }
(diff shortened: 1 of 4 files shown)


regexident:
Awesome @jsbean, you beat me to it!
Fixing order-preservation is what all my recently merged PRs were working up to! 🤓

Nitpick on #17: It's not the order attributes that's unsupported, it's the order of elements.

The XML specification explicitly defines the order of attributes to be of no importance:

> The Name in the start- and end-tags gives the element's type. [Definition: The Name-AttValue pairs are referred to as the attribute specifications of the element], [Definition: with the Name in each pair referred to as the attribute name ] and [Definition: the content of the AttValue (the text between the ' or " delimiters) as the attribute value.] **Note that the order of attribute specifications in a start-tag or empty-element tag is not significant.**  (emphasis mine) [Source](https://www.w3.org/TR/xml/#sec-starttags)

regexident:
> Interesting to hear your comments on this, Vincent.

There's gonna be quite a bit of work ahead of us for making XMLCoder preserve order, but adding an `OrderedDictionary` like in this PR is a great start, imho!

regexident:
Having spent some thoughts on the issue of order-preservation and mixed-content elements in XML I've come to wonder if a `(Ordered)Dictionary` is correct to begin with. 🤔 

### Current `MultiMap` Representation

Currently we use something like `var elements: [String: [Element]]` to encode what effectively is a `MultiMap` (i.e. a map with multiple ordered values per key).
Making a `MultiMap` ordered doesn't make much sense, imho, as it wouldn't really preserve the order in a way we would need it to. 

### New `Array` Representation

What we'd probably want instead is to use a `var elements: [Element]` instead (we can drop the key as `Element` itself holds its key, too) and then do this …

### Element lookup

```swift
subscript(key: Key) -> [Element] {
    get {
        return elements.filter { $0.key == key }
    }
}
```

… for key-wise element lookup.

### Lookup Optimization

If we wanted to make sure we're not regression too much in regards to performance for elements with loooooots of children we could add a key-based index à la `var indexesByKey: [String: IndexSet]` at a reasonable memory overhead.

### Mixed content

This would allow us to also support formats with mixed-content elements …

```xml
<something attribute="42">
    <name>Some name</name>
    Some string
    <![CDATA[Some data]]>
    Some more string
</something>
```

… which is also somewhat related to https://github.com/MaxDesiatov/XMLCoder/issues/12, which tries to encode the same, but using an attribute instead of an element for `.name`.

### Richer Structure Representation

Assuming we would be storing all the content, including the `Some value` span …

```swift
enum Content {
    case string(String)
    case data(Data)
    case element(Element)
}
```

… and then having `var contents: [Content]` instead of `var elements: [Element]`.

So that the above XML would be parsed as something like this:

```swift
var element = Element(
    key: "something",
    attributes: ["attribute": "42"],
    contents: [
        .element(Element(
            key: "name",
            contents: [
                .string("Some name")
            ]
        )),
        .string("Some string"),
        .data("Some data".data(encoding: .utf8)
        .string("Some more string")
    ]
)
```

All in all we're currently grossly oversimplifying the internal data representation, limiting us to a rather narrow set os possible supported encodings.

### Possible usecases

Like with the current data representation we could never encode/decode to/from XHTML, which would be an interesting use of `Codable`. (Whether it's a good idea to use `Codable` for XHTML generation is a whole different question. But I'd like to be able to do it with **XMLCoder** in principle.)

### Attributes everywhere

We also shouldn't restrict the use of attributes to `KeyedBox` (which currently is preventing https://github.com/MaxDesiatov/XMLCoder/issues/12 among other bugs), but allow them for arbitrary boxes. But I digress (and will create a separate discussion issue with a solution proposal (teaser: via composition with `Attributed<T>`) for this at another time, soon-ish, I hope).

jsbean:
@regexident and @MaxDesiatov, thanks for the writeup and the consideration! All sounds fair to me.

Let me know if there are any projects you'd like a hand on.

MaxDesiatov:
Hi @jsbean, would be interesting to hear your thoughts on #70 and #71, there are quite substantial changes proposed and interesting to know what the community thinks about this direction.

MaxDesiatov:
Ordered encoding is now available in `master`, which builds on previous changes implemented by @regexident. As far as I understand this PR is no longer relevant, but please feel free to reopen or create a new PR if needed.

PR #39, opened 2018-12-23
Unfold `guard … else` blocks to allow for setting of breakpoints within

regexident:


regexident:
Despite what codecov claims this is purely a syntactic change.
Code (instruction!) coverage hasn't been changed one bit by this PR.

PR #43, opened 2018-12-24
Improve code coverage of auxiliary types

regexident:


Sources/XMLCoder/Auxiliaries/XMLElement.swift
```diff
@@ -58,9 +58,7 @@ struct _XMLElement {
             }
         }
 
-        elements = Dictionary(elementsByKey) { existingElements, newElements in
-            existingElements + newElements
-        }
+        elements = Dictionary(uniqueKeysWithValues: elementsByKey)
```
MaxDesiatov: from what I see this new code has a different meaning: values with matching keys will be overwritten with new values, while in the old version these values are merged. Is this something we'd definitely like to do?

regexident: The keys are retrieved from another `KeyedBox` via `box.elements.map { key, box in … }`.
No additional keys are being introduced. As such the keys remain unique. :)

PR #46, opened 2018-12-28
Add Optional Error Context 

### LINKED ISSUE #3: Correctly handle parsing errors in `XMLStackParser`
Currently there's a single use of `print` in the codebase for printing an error [here](https://github.com/MaxDesiatov/XMLCoder/blob/master/Sources/XMLCoder/XMLStackParser.swift#L334). I'm not sure that the error is correctly propagated outside to users of `XMLCoder`, this needs to be changed to correctly use `throw`.

### LINKED ISSUE #13: Report error context with a customisable length
As reported in MaxDesiatov/CoreXLSX#12 there are multiple issues where it's hard to understand why parsing of a certain XML has failed. `XMLCoder` should be able to report a snippet of a given XML of specified length around the position where the error has happened. This should be an property of type `Int?` named `errorContextLength` on `XMLDecoder` instance. An error thrown from XML parser that has line/column location should be repackaged having a context around that location of specified length.

For example, if an error was thrown indicating that there's an unexpected character at line 3, column 15 with `errorContextLength` set to `10`, a new error type should be rethrown containing 5 characters before column 15 and 5 characters after, all on line 3. Line wrapping should be handled correctly too as the context can span more than a few lines.

hodovani:
Add optional length of error context.

Resolves #13 
Resolves #3


regexident:
FYI this crashes if `errorPosition - offset < 0` or `errorPosition + offset > string.length`.

MaxDesiatov:
Thanks for reporting @regexident! @hodovani could you please have a look and add corresponding test cases and fixes? Thanks

PR #53, opened 2019-01-02
Remove unnecessary use of `@available(…)` for `OutputFormatting.sortedKeys`

regexident:


MaxDesiatov:
@regexident please don't consider the previous comment as discouraging against substantial improvements, it only means that those will need more time to review and discuss, and I will get to those in the next few weeks anyway.

PR #54, opened 2019-01-02
Fix `KeyDecodingStrategy.custom` signature

MaxDesiatov:
When trying to provide test coverage for `KeyDecodingStrategy.custom` we've stumbled upon an API, which doesn't make much sense. The [corresponding case for JSONDecoder](https://developer.apple.com/documentation/foundation/jsondecoder/keydecodingstrategy/custom) involves creation of a dummy `AnyKey` structure to return a new synthesised coding key, which isn't guaranteed to match a coding key of the decoded type. It's also unclear how `intValue` of this dummy structure could be used.

All other cases of `KeyDecodingStrategy` delegate to functions with `String -> String` signature and my opinion this is the main application of `KeyDecodingStrategy.custom`: convert string keys to other string keys with a supplied transformation.

This is a breaking change though, so I'm still a bit unsure whether we should follow `JSONDecoder` or rely on common sense with this.

@regexident could you please provide your feedback, maybe you had some experience with this? Thanks.

### DIFF
--- a/Sources/XMLCoder/Decoder/XMLDecoder.swift
+++ b/Sources/XMLCoder/Decoder/XMLDecoder.swift
@@ -137,9 +137,9 @@ open class XMLDecoder {
         case convertFromCapitalized
 
         /// Provide a custom conversion from the key in the encoded XML to the keys specified by the decoded types.
-        /// The full path to the current decoding position is provided for context (in case you need to locate this key within the payload). The returned key is used in place of the last component in the coding path before decoding.
+        /// The returned key is used in place of the last component in the coding path before decoding.
         /// If the result of the conversion is a duplicate key, then only one box will be present in the container for the type to decode from.
-        case custom((_ codingPath: [CodingKey]) -> CodingKey)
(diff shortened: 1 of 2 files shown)


regexident:
Hah, if it wasn't for `.custom` (and the exact issue you're describing) the whole key transformation madness within [`XMLKeyedDecodingContainer`](https://github.com/MaxDesiatov/XMLCoder/blob/d4e7673a96b95405991b02e1ca22a53294e687c7/Sources/XMLCoder/Decoder/XMLKeyedDecodingContainer.swift#L32-L66) could removed and the transformations be applied right within `XMLParserDelegate`'s `func parser(_:didStartElement:namespaceURI:qualifiedName:attributes:)`.

I stumbled upon this today, trying to remove all those unnecessary allocations caused by the late transformation. 

There are quite a few places where `Codable`'s API has made some _sub-optimal_ choices.
For the sake of user-friendlyness though I guess we should try to keep the public API as close to JSONCoding, as possible. :/

MaxDesiatov:
thanks @regexident, that's super helpful!

PR #61, opened 2019-01-07
Add test to case when error context size goes outside content size

### LINKED ISSUE #13: Report error context with a customisable length

hodovani:
Resolves #13

### DIFF
--- a/Tests/XMLCoderTests/ErrorContextTests.swift
+++ b/Tests/XMLCoderTests/ErrorContextTests.swift
@@ -1,0 +1,8 @@
+//
+//  ErrorContextTests.swift
+//  XMLCoder
+//
+//  Created by Matvii Hodovaniuk on 1/7/19.
+//
+
+import Foundation
--- a/Tests/XMLCoderTests/ErrorContextTests.swift
+++ b/Tests/XMLCoderTests/ErrorContextTests.swift
@@ -1,8 +1,0 @@
-//
-//  ErrorContextTests.swift
-//  XMLCoder
-//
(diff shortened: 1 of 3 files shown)


Tests/XMLCoderTests/ErrorContextTests.swift
```diff
@@ -0,0 +1,8 @@
+//
+//  ErrorContextTests.swift
+//  XMLCoder
+//
+//  Created by Matvii Hodovaniuk on 1/7/19.
+//
+
+import Foundation
```
MaxDesiatov: was this mostly file committed by mistake?

regexident:
LGTM, too. 🙂

Tests/XMLCoderTests/ErrorContextTests.swift
hodovani: @MaxDesiatov yes, I can't see this file through Xcode. It was committed by mistake.

PR #62, opened 2019-01-08
Add tests for keyed and unkeyed int types

hodovani:


Tests/XMLCoderTests/Minimal/UnkeyedIntTests.swift
```diff
@@ -0,0 +1,261 @@
...
+    """
+    <container>
+        <value>42</value>
+        <value>43</value>
+        <value>44</value>
+    </container>
+    """
+
+class UnkeyedIntTests: XCTestCase {
+    func testInt8() throws {
```
MaxDesiatov: could we create a new `testIntArray` generic function here and call that function with a metatype argument similar to what is available in `KeyedIntTests`?

PR #64, opened 2019-01-10
Remove concrete encode/decode functions

hodovani:
I removed unnecessery typed encode/decode functions in Keyed and Unkeyed Encoder and Decoder. It seems to me that generic functions work well and we don't need to have these ones.

@regexident could you please comment this solution.


MaxDesiatov:
This is a fantastic spot, thanks @hodovani! it never occurred to me that these functions might have a default implementation. It also looks like this on average slightly improves performance 🤩

screen shot 2019-01-10 at 12 34 08

PR #66, opened 2019-01-18
Add single array element example to tests

### LINKED ISSUE #65: Arrays of one element encode ok but fail to decode
With a simple struct:
`struct ProudParent : Codable {`
`       var myChildAge: [Int]`
`}`

`var theparent = ProudParent()`
`theparent.myChildAge = [2,3]`

`<ProudParent><myChildAge>2</myChildAge><myChildAge>3</myChildAge></ProudParent>`

The encoding _always_ works, but if there is only **one** element in the myChildrenAgesCollection Int array it will fail to decode with the error: "Expected to decode Array<Any> but found a string/data instead".

EX: this will encode but fail to decode back
`var theparent = ProudParent()`
`theparent.myChildAge = [2]`

`<ProudParent><myChildAge>2</myChildAge></ProudParent>`

Surely something as simple as that structure can be decoded. What trick am I missing?

MaxDesiatov:
Resolves #65


stonedauwg:
You like those @hodovani? He got those from my issue 😋

PR #73, opened 2019-02-06
Attributed Intrinsic

### LINKED ISSUE #12: Can’t reach an XML value

JoeMatt:
## Overview

This PR fixes #12: decoding of unkeyed single value elements that contain attributes as reported in that issue.

## Example

```xml
<?xml version="1.0" encoding="UTF-8"?>
<foo id="123">456</foo>
``` 

```swift
private struct Foo: Codable, DynamicNodeEncoding {
    let id: String
    let value: String

        case id
        case value
        // case value = "" would also work
    }

    static func nodeEncoding(forKey key: CodingKey) -> XMLEncoder.NodeEncoding {
        switch key {
        case CodingKeys.id:
            return .attribute
        default:
            return .element
        }
    }
}
```

Previously this XML example would fail to decode. This PR allows two different methods of decoding discussed in usage.

## Usage

This PR will support decoding the example XML in two cases as long as the prerequisite cases are matched.

### Prerequisites

1. No keyed child elements exist
2. Any keyed child nodes are supported Attribute types and are indicated to decode as such

### Supported cases

1. An instance var with the key `value` of any decodable type.
2. An instance var of any key that has a Decoding key of String value "value" or "".

The decoder will look for the case where an element was keyed with either "value" or "", but not both, and only one of those values (ie; no other keyed elements). It will automatically find the correct value based on the CodingKey supplied.

## Other considerations

The choice to decode as either "value" or "" keys was purely to try to support the inverse to XML version which would only work if an instance var specifically has a `CodingKey` with associated value type `String` that returns an empty string, if PR #70 is commited as-is, which adds XML coding support for unkeyed attributed value elements.

The 'value' variant was added as a simpler means to support decoding a nested unkeyed element without having to provide custom CodingKey enum for a struct. Something needed to be provided since Swift doesn't have empty string iVars `let "" : String`, isn't a valid iVar token for example, so `value` was chosen as a logical default.

## Notes

This PR is an extension of #70 , though it could be recoded to work off of `master`. The last commit in this PR is the only commit specific to this feature, though #70 provides the inverse solution of creating XML from an attributed value wrapping struct.

Coding and decoding unit tests of String and Int values are included.


JoeMatt:
It's incredibly difficult to find the failing error i the Travis logs. I'd suggest turning off the verbosity, and if linting is going to be a failure it might be faster to incorporate it as a pre-commit hook or use Danger or something that's faster than a full travis build for non-compilation related failures.

MaxDesiatov:
Thanks @JoeMatt, I've fixed SwiftFormat complaints and also merged `master` after #70 was merged. This still causes `KeyedTests` and `UnkeyedTests` to fail though

JoeMatt:
@MaxDesiatov Interesting, I don't think they were failing on my branch. I'll try to take a look this week. I'm moving this code into my own SOAP Swift Coding library that depends on XMLCoder so I'll have to fix any issues that arise. I'll push up fixes as I find them.

MaxDesiatov:
@JoeMatt while I've tried to remove that `DynamicNodeEncoding` extension in one of the commits, I did have to add it back to make all of the tests success before merging #70. The extension is also [available in the corresponding branch](https://github.com/salido/XMLCoder/blob/feature/12-attributedIntrinsic/Sources/XMLCoder/Encoder/DynamicNodeEncoding.swift) of this PR, so doesn't look like that's the reason

thecb4:
Any news here? I'm definitely interesting in using this to support AppStore Connect uploading.

JoeMatt:
A few days I’ll be back on this. Need to fix something in another dependency of my app first.

JoeMatt:
I'm testing / fixing a case where the attributes are enums with associated values that can be reduced to a simple box.

Also, it's ready when it's ready, no need to keep posting for updates. Subscribe to the issue to get notified on updates.

MaxDesiatov:
Thanks @JoeMatt, no worries! I was thinking if I should pick up the branch myself, I wonder if any of the commenters were interested in contributing or planning their own work around this given that it's a pressing issue. Please let me know if there is any way I can help.

MaxDesiatov:
hm, most of the linter warnings reported by @houndci-bot are not valid, it probably didn't pick up the actual config. Sorry about that, it was setup just recently. I'll resolve irrelevant warnings

MaxDesiatov:
also, merged `master`, maybe that'll make it pick up the correct config

JoeMatt:
@MaxDesiatov I have one more thing I'm working on. Doesn't need to be in this PR. 
It's when an enum is the attribute that should be decoded as a simple type, string etc.

example:

```xml
    <number type="unit">A201</number>
```

```swift
   public enum PostingType : Equatable, Codable {
            case unit
            case resno
        }

            let values = try decoder.container(keyedBy: CodingKeys.self)
            if let unit = try values.decodeIfPresent(String.self, forKey: .unit) {
                self = .unit(unit)
                return
            } else {
                let resno = try values.decode(String.self, forKey: .resno)
                self = .resno(resno)
                return
            }
        }

        public func encode(to encoder: Encoder) throws {
            var container = encoder.container(keyedBy: CodingKeys.self)
            switch self {
            case .unit(let unit):
                try container.encode(unit, forKey: .unit)
            case .resno(let resno):
                try container.encode(resno, forKey: .resno)
            }
        }

        case unit(String)
        case resno(String)
    }
    
    public struct PostingNumber : VendorDataProtocol, DynamicNodeEncoding {

        public let type : PostingType

        public init(type: PostingType) {
            self.type = type
        }

        enum CodingKeys : String, CodingKey {
            case type
            case typeValue = ""
        }

        public static func nodeEncoding(forKey key: CodingKey) -> XMLEncoder.NodeEncoding {
            switch key {
            case PostingNumber.CodingKeys.type: return .attribute
            default: return .element
            }
        }

            let type = try container.decodeIfPresent(PostingType.self, forKey: .type)
            let typeArray = try container.decodeIfPresent([PostingType].self, forKey: .type) // Deal with the double value of type as attribute and element
            if let type = type {
                self.type = type
            } else if let typeArray = typeArray, let first = typeArray.first {
                self.type = first
            } else {
                throw DecodingError.keyNotFound(CodingKeys.type, DecodingError.Context.init(codingPath: decoder.codingPath, debugDescription: "Missing type"))
            }
        }

        public func encode(to encoder: Encoder) throws {
            var container = encoder.container(keyedBy: CodingKeys.self)
            switch type {
            case .unit(let unit):
                try container.encode("unit", forKey: .type)
                try container.encode(unit, forKey: .typeValue)
            case .resno(let resno):
                try container.encode("resno", forKey: .type)
                try container.encode(resno, forKey: .typeValue)
            }
        }
    }
```

MaxDesiatov:
looks like all checks pass, thank you very much @JoeMat! I'll review this in the next few days

JoeMatt:
I pushed one more commit which is the codable test for associated type unkeyed intrinsics.

The decoding I couldn't get to work. Requires more knowledge of the decoding code. I tried adding deifference decoders with code like ``

JoeMatt:
Cool, thanks a lot @MaxDesiatov .

Sorry for leaving in those prints. I have a habit of putting them everywhere in my unit tests but in my local setup, they get shuffled to a console logging app. 

XMLCoder is really coming along!

If I see anything else in my code that could be of use I'll definitely make more PRs.

PR #96, opened 2019-05-02
Fixed a bug when decoding a key with one character only

TheFlow95:
I found an issue with the `convertFromCapitalized` key decoding strategy which threw an error when the key had just one characted. This PR should fix the issue.


MaxDesiatov:
Thank you @TheFlow95, great catch! I've also added a test that reproduces the issue.

PR #101, opened 2019-05-11
Always encode empty elements as KeyedBox

MaxDesiatov:
This fixes decoding of empty elements in certain cases that weren't previously covered, specifically MaxDesiatov/CoreXLSX#64.


grin:
@MaxDesiatov a quick question, is https://github.com/MaxDesiatov/XMLCoder/pull/102/commits/237cc28f11f4506b60e30af28f4cb076e30ee799 required for this to work? I'm getting a bunch of non-trivial conflicts when trying to merge it:
Zrzut ekranu 2019-05-16 o 12 50 25

also, it doesn't seem to be working without it, at least for me.

MaxDesiatov:
@grin you're right, this PR is blocked by #102, but some of the tests are still failing, I hope to get to it some time next week. Thanks for your comment, I've just added the "work in progress" label

MaxDesiatov:
As was highlighted in #103 a nice workaround is to make the array element optional. While this PR increases test coverage by deleting `NullBox`, the benchmarks show that for corresponding cases performance degrades by at least 30%. I personally would prefer to increase the test coverage by adding more tests that don't kill the performance, rather than deleting code that makes the code more performant. Closing this and I hope we can find a better solution in the future.

Issue #108, opened 2019-06-26
The order of the attributes changes randomly


acecilia:
I noticed that when there are multiple attributes, every time I encode the model they are ordered in a different way in the resulting `xml`. This is a huge problem during unit tests, where it becomes impossible to validate the encoded `Data` (or `String`)  against a reference one.

Is this a known issue?

Thanks,
Andres

MaxDesiatov:
Hi @acecilia, thank you for reporting this. I thought this was fixed in 0.6.0 with #102 being merged, but interesting to see this still being reproduced. Do you have any specific unit tests you'd be able to share? If so, would you mind if I add them to XMLCoder's test suite?

acecilia:
Sure. Find in https://github.com/MaxDesiatov/XMLCoder/pull/109 some tests that showcase the problem and a small explanation of why this happens. Find in https://github.com/MaxDesiatov/XMLCoder/pull/110 a fix for it :)

PR #109, opened 2019-06-30
Add tests for random attribute order

acecilia:
This is PR to showcase the issue https://github.com/MaxDesiatov/XMLCoder/issues/108. Build should fail due to the `DynamicNodeEncodingTest.testEncode()` test not passing.

The attributes order is random is because a Swift `Dictionary` is used to store them. A `Dictionary` is an unordered collection of keys and values. 

In theory, that is fine, as the XML specification says that the order of the attributes is not important. 

In practise, the result of this is that the encoding operation is not idempotent: despite using the same model, every time an encoding operation is performed, the resulting XML string has different attributes sorting. This is a major problem in several scenarios, but the most important is during tests.

### DIFF
--- a/Tests/XMLCoderTests/DynamicNodeDecodingTest.swift
+++ b/Tests/XMLCoderTests/DynamicNodeDecodingTest.swift
@@ -21,14 +21,18 @@ private let overlappingKeys = """
 private let libraryXMLYN = """
 <?xml version="1.0" encoding="UTF-8"?>
 <library count="2">
-    <book id="123">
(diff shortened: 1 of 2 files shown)


acecilia:
Feel free to close this PR when you do not need it anymore

MaxDesiatov:
@acecilia thank you! Can I assume that this is included in #110?

acecilia:
@MaxDesiatov Yes, this is included in https://github.com/MaxDesiatov/XMLCoder/pull/110

PR #117, opened 2019-07-22
Add Linux support

drewag:
Only a few minor changes needed to be made for this to work properly on Linux.

- Upgraded to Swift 5.0.1
- There was a logical error in XMLStackParser causing a crash on forced unwrapping
- Element names returned from the XMLParser on linux have the namespace coming after the name. I put a hack in to reverse them back.
- The allTests definitions are no longer relevant. Instead, I used `swift test --generate-linuxmain`


.gitignore
```diff
@@ -25,6 +26,7 @@ xcuserdata/
 *.moved-aside
 *.xccheckout
 *.xcscmblueprint
+XMLCoder.xcodeproj
```
MaxDesiatov: I don't think this file can be removed from the repository or added to .gitignore, since it is required for Carthage support.

Sources/XMLCoder/Auxiliaries/XMLStackParser.swift
```diff
@@ -126,6 +126,10 @@ extension XMLStackParser: XMLParserDelegate {
                 namespaceURI: String?,
                 qualifiedName: String?,
                 attributes attributeDict: [String: String] = [:]) {
+        #if os(Linux)
+            // For some reason, element names on linux are coming out with the namespace after the name
+            let elementName = elementName.components(separatedBy: ":").reversed().joined(separator: ":")
```
MaxDesiatov: this is most probably caused by [this issue](https://bugs.swift.org/browse/SR-10424). I wonder if this breaks any of the tests we have for the namespaces, since `XMLDecoder` has a special boolean flag for namespace handling.

Tests/XMLCoderTests/ErrorContextTest.swift
```diff
@@ -29,11 +29,19 @@ final class ErrorContextTest: XCTestCase {
...
+                \(underlying.localizedDescription) \
+                at line 1, column 7:
+                `ah //>`
+                """)
+            #else
+                XCTAssertEqual(ctx.debugDescription, """
+                \(underlying.localizedDescription) \
+                at line 1, column 2:
+                `<blah `
+                """)
```
MaxDesiatov: I don't think this special handling on Linux would be very reliable. I'd definitely prefer this inconsistency to be reported on bugs.swift.org before merging this PR. Links to those bug reports should be added as comments in relevant code and documentation, so that both users and developers are aware of this.

XMLCoder.xcodeproj/project.pbxproj
```diff
@@ -804,7 +804,7 @@
 				PRODUCT_NAME = "$(TARGET_NAME:c99extidentifier)";
 				SKIP_INSTALL = YES;
 				SWIFT_ACTIVE_COMPILATION_CONDITIONS = "$(inherited)";
-				SWIFT_VERSION = 4.2;
+				SWIFT_VERSION = 5.0;
```
MaxDesiatov: is there a specific reason for changing this in the Xcode project file if the PR is only concerned with the Linux support? This change breaks our compatibility with Xcode 10.0, which we'd like to maintain as long as possible.

drewag:
Thanks for the feedback. Can I just confirm that supporting Linux moving forward is a goal worth pursuing before I make these changes?

MaxDesiatov:
@drewag yes, absolutely, I'd be super happy to merge and maintain Linux support, as long as we can see a clear path towards unifying the behaviour across all platforms. I previously did plan to add support for Linux to XMLCoder, but I wasn't sure if Foundation had the feature parity on Linux, especially in terms of XML namespaces support. I'm fine with having small inconsistencies in `master` and maybe even tagged versions, but this has to be clearly documented and tested on CI.

drewag:
Ok great, I'll work on it.

Sources/XMLCoder/Auxiliaries/XMLStackParser.swift
drewag: All test were passing when I tried it. Do we need additional tests to make sure something else specific isn't broken?

XMLCoder.xcodeproj/project.pbxproj
drewag: I had upgraded to Swift 5 to see if that would fix any of the linux errors. But I don't think it is necessary, I'll revert all changes relative to the Swift version

Sources/XMLCoder/Auxiliaries/XMLStackParser.swift
MaxDesiatov: sorry, it seems I previously missed a test case [here](https://github.com/MaxDesiatov/XMLCoder/blob/e17d335683faefa556e820ba0423c3493c2d35ec/Tests/XMLCoderTests/NamespaceTest.swift#L194). We need an additional test where namespaces are present in the source XML, but `shouldProcessNamespaces` is `false`. This can be a copy of the existing test function, but with that property flipped and an assert verifying that the namespace part isn't stripped after decoding.

MaxDesiatov: another thing in this new namespace test: you'd most probably need to preserve the namespace in coding keys, which probably requires copying the decoded type and adding the namespaces there in the coding keys of the copy. Hope this makes sense, please let me know if there are any questions.

drewag:
There are two failing tests on Linux if I revert to Swift 4.2. Looking into what the problems are now.

drewag:
One failed test is NoteTest.testInvalidXML. The 4.2 parser fails to catch the "Ffrom" inconsistency. Not sure what to do about that except ignore that test on versions before 5.0.

The harder one is ErrorContextTest.testErrorContext.

There we are getting the wrong error type:

> typeMismatch(Swift.Dictionary<Swift.String, Any>, Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "value", intValue: nil)], debugDescription: "Expected to decode Dictionary<String, Any> but found SharedBox<Array<Box>> instead.", underlyingError: nil))

That one is a bit above my pay grade. Any ideas?

MaxDesiatov:
Are both of those errors reproducible on 4.2, but only on Linux? If so, we could keep 4.2 compatibility on Apple platforms, but require 5.0 on Linux. This would mean not testing 4.2 Linux on CI and mentioning that Linux support requires 5.0 in `README.md`

drewag:
Yup that's correct, all tests pass on Linux with Swift 5. I'll update the README and then we should be ready to merge.

drewag:
Sorry for all the commits (Azure Pipelines is new to me). This should finally be ready to merge.

PR #132, opened 2019-09-23
Conditionally conform to Combine.TopLevelDecoder

sharplet:
This enables use of `XMLDecoder` with Combine's `decode(_:decoder:)` operator by conditionally conforming to `TopLevelDecoder` when Combine is available.

I explored doing the same for `TopLevelEncoder`, but `XMLEncoder` currently requires `rootKey:` to be specified at encoding time, so it's not possible to provide an implementation of `encode(_:)` without some design around how to dynamically determine the root key to use.

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,25 @@ extension IntOrString: Codable {
 This is described in more details in PR [\#119](https://github.com/MaxDesiatov/XMLCoder/pull/119) 
 by [@jsbean](https://github.com/jsbean) and [@bwetherfield](https://github.com/bwetherfield).
 
+## Integrating with [Combine](https://developer.apple.com/documentation/combine)
+
+When Apple's Combine framework is available, `XMLDecoder` conforms to the
+`TopLevelDecoder` protocol, which allows it to be used with the
+`decode(type:decoder:)` operator:
+
+```swift
+import Combine
+import Foundation
+import XMLCoder
+
+func fetchBook(from url: URL) -> AnyPublisher<Book, Error> {
+    return URLSession.shared.dataTaskPublisher(for: url)
+        .map(\.data)
+        .decode(type: Book.self, decoder: XMLDecoder())
+        .eraseToAnyPublisher()
+}
+```
+
 ## Installation
 
 ### Requirements
--- a/Tests/XMLCoderTests/CombineTests.swift
+++ b/Tests/XMLCoderTests/CombineTests.swift
@@ -1,0 +1,37 @@
+//
+//  CombineTests.swift
+//  XMLCoder
+//
+//  Created by Adam Sharp on 9/29/19.
+//
+
+#if canImport(Combine)
+import Combine
+import Foundation
+import XCTest
+import XMLCoder
+
+private let xml = """
+<?xml version="1.0" encoding="UTF-8"?>
+<foo>
+    <name>Foo</name>
+</foo>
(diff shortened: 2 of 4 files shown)


MaxDesiatov:
@sharplet thank you for the PR! Would you mind adding a short example of use with Combine to README and also a unit test (maybe based on the example code)?

sharplet:
Add a test case and a new section to the README, let me know what you think!

sharplet:
Looks like the swiftpm tests are failing on CI, possibly because the Swift compiler version isn't recent enough to know about the `macCatalyst` platform? I could add a `#if !os(Linux)` conditional to the new test case, which feels a little clunky but should work.

sharplet:
One other option would be to just drop `macCatalyst` from the availability annotation. I've pushed a commit which explicitly excludes Linux, in case you're happy with that option.

sharplet:
`!os(Linux)` didn't work, and I released that because Catalyst is brand new there should always be a version of Combine available. 🤞

sharplet:
Good news: `CombineTests` is actually running now on the Xcode 11 job, but it's crashing, which might be because the job is linking against the macOS SDK, but is running on 10.14 where Combine.framework isn't available at runtime. That test run might need to use the iOS SDK in order to successfully link against Combine.

Tests/XMLCoderTests/CombineTests.swift
```diff
@@ -0,0 +1,37 @@
...
+<foo>
+    <name>Foo</name>
+</foo>
+""".data(using: .utf8)!
+
+private struct Foo: Decodable {
+    var name: String
+}
+
+@available(iOS 13.0, macOS 10.15.0, tvOS 13.0, watchOS 6.0, *)
```
MaxDesiatov: I wonder if the solution would be to remove `macOS 10.15.0` clause until that's released? We would avoid crashing on CI until then too. After all I wouldn't mind tagging a minor release just for enabling that later when Catalina is released and available on CI.

README.md
```diff
@@ -269,6 +269,25 @@ extension IntOrString: Codable {
...
+`decode(type:decoder:)` operator:
+
+```swift
+import Combine
+import Foundation
+import XMLCoder
+
+func fetchBook(from url: URL) -> AnyPublisher<Book, Error> {
+    return URLSession.shared.dataTaskPublisher(for: url)
+        .map(\.data)
```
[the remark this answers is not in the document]
sharplet: This is actually an explicit overload that Combine provides: https://developer.apple.com/documentation/combine/publisher/3343975-map

Tests/XMLCoderTests/CombineTests.swift
sharplet: Removing macOS from the availability annotation would instead result in a compiler error. The issue is that we successfully compile & link, but then attempt to run the tests on a platform where the framework doesn't exist and instead crash when the symbol can't be loaded.

If we want to exclude the test from macOS altogether, we could add `#if !os(macOS)` to the compiler directive at the top of the file. The tests won't be exercised on CI then, so it might still make sense to add a CI job that runs on Xcode 11 + iOS?

MaxDesiatov: The Xcode 11 job already tests on macOS, iOS and tvOS, so that should be fine.

PR #145, opened 2019-10-16
Fix Decoding of Empty String

bwetherfield:
Prior to this fix, we found that decoding an empty string (represented as a null element), `<string/>` was confusing the decoder. I can't seem to find any issues logged. Perhaps @jsbean has an idea where these might be hiding...?

I'm also building up to a PR that addresses the array-of-empty-elements issue, but I am trying to disentangle the changes we merged into a separate branch!


bwetherfield:
This one is now ready for your review, @MaxDesiatov ! Thanks

Tests/XMLCoderTests/Minimal/StringTests.swift
```diff
@@ -25,15 +25,6 @@ class StringTests: XCTestCase {
...
 
-    func testMissing() {
-        let decoder = XMLDecoder()
-
-        let xmlString = "<container />"
-        let xmlData = xmlString.data(using: .utf8)!
-
-        XCTAssertThrowsError(try decoder.decode(Container.self, from: xmlData))
-    }
-
```
MaxDesiatov: Does this test pass? If so, I'd prefer to keep it in the test suite

bwetherfield: I've played around with this, and it does work if the element is renamed to, say, `notValue`, but I'm encountering an ambiguity with the treatment of the `value` intrinsic. Would you be content with this if we pushed through the proposed change from `value` to `xmlValue` as the labeling of the intrinsic?

bwetherfield: (or even limiting the intrinsic to `""`)

MaxDesiatov: `""` and `"xmlValue"` are both fine, `"xmlValue"` (or everything that starts with "xml" for that matter) is not a valid element or attribute, at least according to my understanding of the XML standard:

>Names beginning with the string "xml", or with any string which would match (('X'|'x') ('M'|'m') ('L'|'l')), are reserved for standardization in this or future versions of this specification.

MaxDesiatov: On the other hand, isn't decoding the value of `<container />` element to an empty string is the expected behaviour here? I think I also relied on this in [CoreXLSX](https://github.com/MaxDesiatov/CoreXLX) in some of its behavior if I remember correctly, but this needs to be confirmed.

MaxDesiatov: This might seem a bit self-centered, but if it doesn't break CoreXLSX, I'm probably fine with the change, hope MusicXML test suite would pass as well. Just wondering how many other apps and libraries we could break with this change and what would be possible workarounds in that case.

bwetherfield: Hoping we can perhaps circumvent these concerns by changing the use of the `"value"` coding key, per my latest PR! It does mean downstream users would have to introduce `= ""` additions, but since this change is probably on the way anyway, I wonder if now is as good a time as any...?

MaxDesiatov:
@bwetherfield I'm ready to review this again when conflicts are resolved 🙂

bwetherfield:
Thanks, @MaxDesiatov ! I also need to add back the missing test :) Will keep you posted

bwetherfield:
OK - should be ready for review! Thanks, @MaxDesiatov

Tests/XMLCoderTests/Minimal/StringTests.swift
MaxDesiatov: @bwetherfield Will this test be added back?

bwetherfield:
My apologies, @MaxDesiatov - should be in now!

PR #147, opened 2019-10-17
Add whitespace trimming test with copyright symbol

### LINKED ISSUE #141: Value with copyright symbol © has its preceding whitespace trimmed off even `trimValueWhitespaces` is set to false
See test case added in my commit that verifies this issue
https://github.com/MaxDesiatov/XMLCoder/commit/82d9bac8a39926960efdd9a93d54af3ead0d2121

Regardless of nested or not, parsing something like `<t>Copyright © 2019 Company, Inc.</t>` gets me `"Copyright© 2019 Company, Inc" ` with the preceding whitespace before `©` trimmed off.

I look for the codebase and found the only place involving `trimmingCharacters` is in `XMLStackParser:114`. But later I found that the Foundation method to trim whitespace is not the culprit either.

```swift
XCTAssertEqual("Copyright © 2019 Company, Inc.".trimmingCharacters(in: .whitespacesAndNewlines), "Copyright © 2019 Company, Inc.") => true
```

So I am confused. Any pointer is appreciated.

MaxDesiatov:
Resolve #141

### DIFF
--- a/Tests/XMLCoderTests/SpacePreserveTest.swift
+++ b/Tests/XMLCoderTests/SpacePreserveTest.swift
@@ -65,10 +65,12 @@ final class SpacePreserveTest: XCTestCase {
     }
 
     func testCopyRight() throws {
-        let result = try XMLDecoder().decode(String.self, from: copyrightXML)
+        let decoder = XMLDecoder()
+        decoder.trimValueWhitespaces = false
+        let result = try decoder.decode(String.self, from: copyrightXML)
         XCTAssertEqual(result, "Copyright © 2019 Company, Inc.")
 
-        let item = try XMLDecoder().decode(Item.self, from: nestedCopyrightXML)
+        let item = try decoder.decode(Item.self, from: nestedCopyrightXML)
         XCTAssertEqual(item, Item(text: "Copyright © 2019 Company, Inc."))
 
         XCTAssertEqual("Copyright © 2019 Company, Inc.".trimmingCharacters(in: .whitespacesAndNewlines), "Copyright © 2019 Company, Inc.")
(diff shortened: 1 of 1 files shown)


Tests/XMLCoderTests/SpacePreserveTest.swift
```diff
@@ -51,4 +63,14 @@ final class SpacePreserveTest: XCTestCase {
...
     }
+
+    func testCopyRight() throws {
+        let result = try XMLDecoder().decode(String.self, from: copyrightXML)
+        XCTAssertEqual(result, "Copyright © 2019 Company, Inc.")
+
+        let item = try XMLDecoder().decode(Item.self, from: nestedCopyrightXML)
+        XCTAssertEqual(item, Item(text: "Copyright © 2019 Company, Inc."))
+
+        XCTAssertEqual("Copyright © 2019 Company, Inc.".trimmingCharacters(in: .whitespacesAndNewlines), "Copyright © 2019 Company, Inc.")
```
hodovani: @MaxDesiatov can a repeated string be replaced with a constant?

Tests/XMLCoderTests/SpacePreserveTest.swift
```diff
@@ -16,12 +16,24 @@ private let nestedXML = """
 <si><t xml:space="preserve"> </t></si>
 """.data(using: .utf8)!
 
+private let copyrightXML = """
+<t>Copyright © 2019 Company, Inc.</t>
+""".data(using: .utf8)!
+
+private let nestedCopyrightXML = """
+<si><t>Copyright © 2019 Company, Inc.</t></si>
```
hodovani: @MaxDesiatov we might replace a repeated string with a constant here.

Tests/XMLCoderTests/SpacePreserveTest.swift
MaxDesiatov: great point, fixed now

PR #149, opened 2019-11-03
Replace value intrinsic with empty string key

bwetherfield:
## Overview
Restricts the special case usage of `"value"` OR `""` as a coding key, as implemented in #73 to just `""`. `"value"` resumes functioning as a normal coding key. 

## Example
As noted in #145, `"value"` is starting to collide awkwardly in one or two cases where it may be _implicit_ per the special case usage 
```swift
<value-containing>I am implicitly labeled</value-containing>
``` 
or _explicit_, as in 
```swift
<value-containing>
    <value>I am explicitly labeled</value>
</value-containing>
``` 
With the change proposed, the latter example will be unambiguously captured by
```swift
struct ValueContaining: Codable {
    let value: String
}
```
while the former is equivalent to
```swift
struct ValueContaining: Codable {
    let value: String

        case value = ""
    }
}
```
## Source Compatability

This is a **breaking change** for downstream users of the special `"value"` coding key. All such uses should be replaced by `case value = ""` subject to this PR's inclusion.

### DIFF
--- a/Sources/XMLCoder/Auxiliaries/Box/KeyedBox.swift
+++ b/Sources/XMLCoder/Auxiliaries/Box/KeyedBox.swift
@@ -24,7 +24,7 @@ struct KeyedBox {
     }
 
     var value: SimpleBox? {
-        return elements["value"].first as? SimpleBox
+        return elements[""].first as? SimpleBox
     }
 }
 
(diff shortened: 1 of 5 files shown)


bwetherfield:
If this change goes through, and anyone, like [us](https://github.com/dn-m/MusicXML/), find that a bunch of explicit `CodingKey` implementations are now needed (where before you just set-and-forgot a variable `let value: String`, for example), you may find [this template](https://github.com/dn-m/MusicXML/blob/update-value-coding-key-dependency/Sources/MusicXML/Templates/ExplicitCodingKey.stencil) helpful (for use with [Sourcery](https://github.com/krzysztofzablocki/Sourcery)).

MaxDesiatov:
@bwetherfield thanks for the PR!

PR #152, opened 2019-11-14
Fix Decoding of Arrays of Empty Elements

### LINKED ISSUE #123: Decoding containers with (potentially)-empty elements
We have a use case where we need to decode an array of potentially empty elements. 

Currently, unkeyed and keyed containers holding onto empty (e.g., a struct with no fields) or a potentially empty (e.g., a struct with all optional fields) elements does not decode.

Say we've got a struct like this:

```Swift
struct Empty: Equatable, Codable { }
```

If we have some XML like this:

```Swift
let xml = """
<container>
    <empty/>
    <empty/>
    <empty/>
</container>
"""
```

It would be nice to decode like this:

```Swift
let decoded = try XMLDecoder().decode([Empty].self, from: xml.data(using: .utf8)!)
// Throws "Expected Empty but found null instead."
```

The poetics are strong.

Wrapping it up a bit, consider this:

```Swift
struct EmptyArray: Equatable, Codable {
    enum CodingKeys: String, CodingKey { case empties = "empty" }
    let empties: [Empty]
}
```

Given the `xml` from above, we can try to decode it like this:

```Swift
let decoded = try XMLDecoder().decode(EmptyArray.self, from: xml.data(using: .utf8)!)
// Throws "Expected Array<Empty> value but found null instead."
```

Lastly, given this container:

```Swift
struct EmptyWrapper {
    let empty: Empty
}
```

And we have some XML like this:

```Swift
let xml = """
<wrapper>
    <empty/>
</wrapper>
"""
```

And we try to decode it like this:

```Swift
let decoded = try XMLDecoder().decode(EmptyWrapper.self, from: xml.data(using: .utf8)!)
// Throws "Expected Empty value but found null instead"
```

We get the same outcomes if you substitute the `Empty` type with something like this:

```Swift
struct MaybeEmpty {
    var idunno: Int?
}
```

If this seems like something that should theoretically be supported, I can give a go at it. It smells like it might touch similar neighborhoods of the codebase as #122.

bwetherfield:
## Overview

Fixes #123 and extends #145 to accommodate decoding of arrays of empty strings or mixed arrays of non-empty and empty strings.

## Example

We may now decode 
```xml
<container>
    <empty/>
    <empty/>
    <empty/>
</container>
```
into the following type
```swift
struct EmptyArray: Equatable, Codable {
    enum CodingKeys: String, CodingKey { case empties = "empty" }
    let empties: [Empty]
}
```
where 
```swift
struct Empty: Equatable, Codable { }
```
We can also decode a value of the following type
```swift
struct EmptyWrapper {
    let empty: Empty
}
```
from 
```xml
<wrapper>
    <empty/>
</wrapper>
```
Further, following from #145 we can decode arrays of empty strings
```xml
<container>
    <string-type/>
    <string-type>My neighbors are empty<string-type>
    <string-type/>
</container>
```
as
```swift
struct EmptyArray: Equatable, Codable {
    enum CodingKeys: String, CodingKey { case stringType = "string-type" }
    let stringType: [String]
}
```
and variants.
## Source Compatibility

In cases where we decode an array of type `[String?]`, an empty element is now decoded as `""` rather than `nil`, the justification being that `String` can itself take empty values. We use the convention that `nil` signifies the absence of an element (if 0 or 1 of the element are allowed) as in the updated [BreakfastTest](https://github.com/bwetherfield/XMLCoder/blob/0d20614e47df98d1a10174e992c585edf629c9b9/Tests/XMLCoderTests/BreakfastTest.swift) and in the following error-throwing [test case](https://github.com/MaxDesiatov/XMLCoder/blob/2855777ff868e8a4c1d944c7da0ddb49a8b3893e/Tests/XMLCoderTests/Minimal/NullTests.swift#L56-L68).


bwetherfield:
I can't quite figure out if
```swift
    func parser(_: XMLParser,
                didStartElement elementName: String,
                namespaceURI: String?,
                qualifiedName: String?,
                attributes attributeDict: [String: String] = [:])
```
and 
```swift
    func parser(_: XMLParser,
                didEndElement _: String,
                namespaceURI _: String?,
                qualifiedName _: String?)
```
in [`XMLStackParser`](https://github.com/MaxDesiatov/XMLCoder/blob/master/Sources/XMLCoder/Auxiliaries/XMLStackParser.swift) get called differently for these two cases in the `XMLParser` implementation. If so, then maybe we could tweak the parser delegate implementation, but I'm not too sure. AFAIK, XML implementations wouldn't tend to distinguish in this way as a rule, so would this cause compatibility issues more broadly?

MaxDesiatov:
@bwetherfield thank you for polishing this and finding new edge cases! 🙏

PR #154, opened 2019-11-25
Mixed choice/non-choice encoding

bwetherfield:
## Overview

Fixes bug encountered when encoding structs that hold a mixture of choice-element and non-choice-element (or multiple choice-element) properties.

## Example

Given a structure that stores both a choice and non-choice property,
```swift
private struct IntOrStringAndDouble: Equatable {
    let intOrString: IntOrString
    let decimal: Double
}
```

the natural encoding approach (now available) is 
```swift
extension IntOrStringAndDouble: Encodable {
        case decimal
    }

    func encode(to encoder: Encoder) {
        try intOrString.encode(to: encoder)
        var container = encoder.container(keyedBy: CodingKeys.self)
        try container.encode(decimal, forKey: .decimal)
    }
}
```

<details><summary>The following <code>encode</code> implementation also works</summary>
<p>

```swift
extension IntOrStringAndDouble: Encodable {
        case decimal
    }

    func encode(to encoder: Encoder) {
        var singleValueContainer = encoder.singleValueContainer()
        try singleValueContainer.encode(intOrString)
        var container = encoder.container(keyedBy: CodingKeys.self)
        try container.encode(decimal, forKey: .decimal)
    }
}
```

</p>
</details>

<details><summary><code>IntOrString</code> as defined in #119 </summary>
<p>

```swift 
enum IntOrString: Equatable {
    case int(Int)
    case string(String)
}

extension IntOrString: Encodable {
        case int
        case string
    }

    func encode(to encoder: Encoder) throws {
        var container = encoder.container(keyedBy: CodingKeys.self)
        switch self {
        case let .int(value):
            try container.encode(value, forKey: .int)
        case let .string(value):
            try container.encode(value, forKey: .string)
        }
    }
}

extension IntOrString.CodingKeys: XMLChoiceCodingKey {} // signifies that `IntOrString` is a choice element
```

</p>
</details>

## Implementation Details

In cases where choice and non-choice elements (or multiple choice elements) co-exist in a keyed container, we merge them into a single `XMLKeyedEncodingContainer` (wrapping a `SharedBox<KeyedBox>`). Arrays of choice elements (using `XMLUnkeyedEncodingContainer` under the hood) are encoded the same way as before, as we do not hit the merging cases. For the array case, we still need the `XMLChoiceEncodingContainer` structure.

## Source Compatibility

This is an additive change.


Sources/XMLCoder/Encoder/XMLEncoderImplementation.swift
```diff
@@ -73,47 +73,85 @@ class XMLEncoderImplementation: Encoder {
...
-        let container = XMLKeyedEncodingContainer<Key>(referencing: self, codingPath: codingPath, wrapping: topContainer)
-        return KeyedEncodingContainer(container)
     }
 
     public func choiceContainer<Key>(keyedBy _: Key.Type) -> KeyedEncodingContainer<Key> {
-        let topContainer: SharedBox<ChoiceBox>
         if canEncodeNewValue {
             // We haven't yet pushed a container at this level; do so here.
-            topContainer = storage.pushChoiceContainer()
+            let container = XMLChoiceEncodingContainer<Key>(
```
MaxDesiatov: likewise here

Sources/XMLCoder/Encoder/XMLEncoderImplementation.swift
```diff
@@ -73,47 +73,85 @@ class XMLEncoderImplementation: Encoder {
 
     public func keyedContainer<Key>(keyedBy _: Key.Type) -> KeyedEncodingContainer<Key> {
         // If an existing keyed container was already requested, return that one.
-        let topContainer: SharedBox<KeyedBox>
         if canEncodeNewValue {
             // We haven't yet pushed a container at this level; do so here.
-            topContainer = storage.pushKeyedContainer()
+            let container = XMLKeyedEncodingContainer<Key>(
```
MaxDesiatov: the comment above seems to be outdated now

Sources/XMLCoder/Encoder/XMLEncoderImplementation.swift
```diff
@@ -73,47 +73,85 @@ class XMLEncoderImplementation: Encoder {
...
+                return KeyedEncodingContainer(container)
+            default:
                 preconditionFailure(
                     """
-                    Attempt to push new (single element) keyed encoding container when already \
-                    previously encoded at this path.
+                    Attempt to push new keyed encoding container when already previously encoded \
+                    at this path.
                     """
                 )
```
MaxDesiatov: Overall this `switch` seems to be a duplicate of `switch` in `public func keyedContainer` modified above, could it be generalized somehow?

bwetherfield: Ah, good point!

PR #157, opened 2019-12-17
Support for mixed-content nodes

ultramiraculous:
Adding support for nodes with mixed content, where loose text nodes might coexist with other  node types. An example of this in the tests is: `<container>first<b>bold text</b>second</container>`.

This restructures the contents of nodes, so that raw text is always stored in a `""`-keyed node, which can occur multiple times if necessary. Previously all text nodes were appended together into the element's `value`, so their order relative to elements between them was lost.


MaxDesiatov:
For some reason Azure Pipelines status is not reported to GitHub now, but [the CI job has passed](https://dev.azure.com/max0484/max/_build/results?buildId=455&view=results). I assume the issue with this is intermittent and going to merge it as is.

PR #182, opened 2020-05-20
Make TopLevelEncoder implementation overridable

kkk669:
`TopLevelEncoder` implementation was added by #175. However, its method cannot be overridden. If it can be, we can use custom root key, root attributes, or header even when we use Combine-style like this:

```swift
final class MyEncoder: XMLEncoder {
    override func encode<T>(_ value: T) throws -> Data where T : Encodable {
        try self.encode(value, withRootKey: "foo", rootAttributes: nil, header: XMLHeader(version: 1.0, encoding: "UTF-8"))
    }
}
```

So, I proposed the solution.

### DIFF
--- a/Sources/XMLCoder/Encoder/XMLEncoder.swift
+++ b/Sources/XMLCoder/Encoder/XMLEncoder.swift
@@ -380,7 +380,7 @@ open class XMLEncoder {
     // MARK: - TopLevelEncoder
 
     open func encode<T>(_ value: T) throws -> Data where T: Encodable {
-        try encode(value, withRootKey: nil, rootAttributes: nil, header: nil)
+        return try encode(value, withRootKey: nil, rootAttributes: nil, header: nil)
     }
 }
 
--- a/Sources/XMLCoder/Encoder/XMLEncoder.swift
+++ b/Sources/XMLCoder/Encoder/XMLEncoder.swift
@@ -379,9 +379,11 @@ open class XMLEncoder {
 
     // MARK: - TopLevelEncoder
 
+    #if canImport(Combine) || canImport(OpenCombine)
     open func encode<T>(_ value: T) throws -> Data where T: Encodable {
         return try encode(value, withRootKey: nil, rootAttributes: nil, header: nil)
     }
+    #endif
 }
 
 private extension String {
--- a/Sources/XMLCoder/Encoder/XMLEncoder.swift
+++ b/Sources/XMLCoder/Encoder/XMLEncoder.swift
@@ -376,6 +376,12 @@ open class XMLEncoder {
         return element.toXMLString(with: header, formatting: outputFormatting)
             .data(using: .utf8, allowLossyConversion: true)!
     }
+
+    // MARK: - TopLevelEncoder
+
+    open func encode<T>(_ value: T) throws -> Data where T: Encodable {
+        try encode(value, withRootKey: nil, rootAttributes: nil, header: nil)
+    }
 }
 
 private extension String {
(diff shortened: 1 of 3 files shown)


Sources/XMLCoder/Encoder/XMLEncoder.swift
```diff
@@ -376,6 +376,12 @@ open class XMLEncoder {
         return element.toXMLString(with: header, formatting: outputFormatting)
             .data(using: .utf8, allowLossyConversion: true)!
     }
+
+    // MARK: - TopLevelEncoder
+
+    open func encode<T>(_ value: T) throws -> Data where T: Encodable {
+        try encode(value, withRootKey: nil, rootAttributes: nil, header: nil)
```
MaxDesiatov, line 383:
```suggestion
        return try encode(value, withRootKey: nil, rootAttributes: nil, header: nil)
```

Tests/XMLCoderTests/CombineTests.swift
```diff
@@ -27,6 +27,12 @@ private struct Foo: Codable {
     var name: String
 }
 
+final class CustomEncoder: XMLEncoder {
+    override func encode<T>(_ value: T) throws -> Data where T : Encodable {
+        try self.encode(value, withRootKey: "bar", rootAttributes: nil, header: nil)
```
MaxDesiatov, line 32:
```suggestion
        return try self.encode(value, withRootKey: "bar", rootAttributes: nil, header: nil)
```

Sources/XMLCoder/Encoder/XMLEncoder.swift
kkk669: I fixed in 84c33e77a57cd008c86321476f5e5a506d80fd76.

Tests/XMLCoderTests/CombineTests.swift
kkk669: I fixed in 84c33e77a57cd008c86321476f5e5a506d80fd76.

Tests/XMLCoderTests/CombineTests.swift
```diff
@@ -27,6 +27,12 @@ private struct Foo: Codable {
     var name: String
 }
 
+final class CustomEncoder: XMLEncoder {
+    override func encode<T>(_ value: T) throws -> Data where T : Encodable {
```
MaxDesiatov, line 31:
```suggestion
    override func encode<T>(_ value: T) throws -> Data where T: Encodable {
```

Tests/XMLCoderTests/CombineTests.swift
```diff
@@ -27,6 +27,12 @@ private struct Foo: Codable {
...
+        return try self.encode(value, withRootKey: "bar", rootAttributes: nil, header: nil)
```
MaxDesiatov, line 32:
```suggestion
        return try encode(value, withRootKey: "bar", rootAttributes: nil, header: nil)
```

Tests/XMLCoderTests/CombineTests.swift
kkk669: I'm sorry I didn't check SwiftFormat. I fixed in 18282b5515145500190dd817c9b0eab2b99188ed.

Tests/XMLCoderTests/CombineTests.swift
kkk669: I fixed in 18282b5515145500190dd817c9b0eab2b99188ed.

kkk669:
@MaxDesiatov It seems that [danger-lint has failed](https://github.com/MaxDesiatov/XMLCoder/pull/182/checks?check_run_id=693713732#step:4:8) in `XMLDecoder`, but I didn't change `XMLDecoder` at all. Do you have any ideas?

MaxDesiatov:
No worries, that seems like a SwiftLint bug 🙂

PR #192, opened 2020-06-08
Property wrapper syntax for node encoding/decoding

### LINKED ISSUE #189: Implement @Element and @Attribute property wrappers
Property wrappers could be a great fit to replace the `DynamicNodeEncoding` API. E.g.

```swift
struct Tag: Codable {
  @Element var element: String
  @Attribute var attribute: String
}
```

would be encoded as `<Tag attribute="foo"><element>bar</element></Tag>`.

There's some bike-shedding needed here with the names, I thought about `XMLElement` instead of `Element`, on the other hand it would be easy to disambiguate as `XMLCoder.Element` if needed. Same with `XMLAttribute` vs `Attribute`.

bwetherfield:
Resolves #189


bwetherfield:
Update: I'm leaving in DynamicNodeEncoding for now as a workaround for the one test case that is causing me problems, but otherwise having the encoder use property attributes to determine node types.

Tests/XMLCoderTests/AdvancedFeatures/AttributedIntrinsicTest.swift
```diff
@@ -104,42 +95,24 @@ private struct FooValue: Codable, Equatable {
     }
 }
 
-private struct FooOptional: Codable, DynamicNodeEncoding, Equatable {
```
MaxDesiatov, line 107: I understand that this is work in progress, but it would be great to keep `DynamicNodeEncoding` in at least a few more versions, but to mark it as deprecated. Unfortunately, this means that we'll have to test both implementations at the same time. Maybe by keeping two versions of the same type, one declared with `DynamicNodeEncoding`, the other using property wrappers. This is the best way to make things easier for users of XMLCoder so that they have some time to migrate, while still getting other new features and bugfixes with new versions.

Sources/XMLCoder/Auxiliaries/XMLAttributeNode.swift
```diff
@@ -0,0 +1,29 @@
+//
+//  XMLAttribute.swift
+//  XMLCoder
+//
+//  Created by Benjamin Wetherfield on 6/3/20.
+//
+
+public protocol XMLAttributeProtocol {}
```
MaxDesiatov, line 8:
```suggestion
#if swift(>=5.1)
public protocol XMLAttributeProtocol {}
```

Sources/XMLCoder/Auxiliaries/XMLAttributeNode.swift
```diff
@@ -0,0 +1,29 @@
...
+        try wrappedValue.encode(to: encoder)
+    }
+
+    public init(from decoder: Decoder) throws {
+        try wrappedValue = .init(from: decoder)
+    }
+}
+
+extension XMLAttributeNode: Equatable where Value: Equatable {}
+extension XMLAttributeNode: Hashable where Value: Hashable {}
```
MaxDesiatov, line 29: Would need the same conditionals in the rest of the places where these are used, including tests.

```suggestion
extension XMLAttributeNode: Hashable where Value: Hashable {}
#endif
```

Tests/XMLCoderTests/AdvancedFeatures/AttributedIntrinsicTest.swift
bwetherfield: Totally!

Sources/XMLCoder/Auxiliaries/XMLAttributeNode.swift
bwetherfield: Aaaah, good point!

bwetherfield:
I like the idea of `Both` being default behavior, as that name is not the nicest to look at. In terms of default encoding, we have previously had `.element` as default (in the other implementation). So that's the only question mark. We will have a lot more `@XMLElement`'s to add, as this is pretty common. Not necessarily a bad thing to be explicit that way. 

re the naming, when I name the property wrapper `XMLElement` I get a funny error from the compiler (one of the reasons I stuck `Node` on the end):

`/Users/ben/Documents/Projects/bwetherfield/XMLCoder/Tests/XMLCoderTests/ClassTests.swift:14:5: Unknown attribute 'XMLElement'`

Must be a naming conflict somewhere I think. If I change to `@XMLCoder.XMLElement` the compiler is happy. So maybe we go for `@XMLCoder.Element` and `XMLCoder.Attribute` instead? What do you think, @MaxDesiatov ?

bwetherfield:
I think we are probably getting a name conflict with `XMLElement` [here](https://developer.apple.com/documentation/foundation/xmlelement)

MaxDesiatov:
`@Element` and `@Attribute` LGTM, and you're right, those can be disambiguated in client code as `@XMLCoder.Element` and `@XMLCoder.Attribute` if needed.

With regards to the third option, we have some assymetry here:

```swift
public enum NodeDecoding {
        case attribute
        case element
        case elementOrAttribute
}
```

Interestingly enough `NodeDecoding` doesn't have a `default` static property.

```swift
public enum NodeEncoding {
        case attribute
        case element
        case both

        public static let `default`: NodeEncoding = .element
}
```

Thus `@Both` makes sense only for encoding. I don't have a good suggestion for it, especially with the assymetry. It could be `@BothAttributeAndElement`, which is ugly, but explicit, and maybe then it could behave symmetrically for both encoding and decoding?

bwetherfield:
I've just rolled back the existing `DynamicNodeEncoding` tests. Mostly wanted to see whether things would work quickly. Now we can add separate tests conditional on `swift(>= 5.1)`

bwetherfield:
I am wondering if `.elementAndAttribute` could be a more descriptive name than `.both` (while also serving symmetry a little better). Then `@ElementAndOrAttribute` could be an option, but I don't know how clear that actually is! What do you think?

MaxDesiatov:
Maybe `@ElementAndAttribute`? In the decoding case it would read as "looking for value in both element and attribute" and in the encoding case as "writing the value to both element and attribute". What do you think?

bwetherfield:
Yeah, I was just thinking that. I think that's a great idea!

transat:
Hi @MaxDesiatov. I have an XML full of attributes. Do I need this PR? Any chance for examples? Cheers.

MaxDesiatov:
@transat either with this PR or without it you'll probably have to create two separate model types. In `master` you'll have to use [dynamic node coding protocols](https://github.com/MaxDesiatov/XMLCoder#dynamic-node-coding) for it if that makes sense?

transat:
Thanks @MaxDesiatov Yes, I've managed to successfully decode the attribute-only version of the file and will give the element-only version a go next. 

Could I keep the same model and ostensibly do something like this?

```static func nodeDecoding(for key: CodingKey) -> XMLDecoder.NodeDecoding {
          if ( _file's root element to String == "sound"_ ){
             return .attribute
          }
          else {
             return .element
          }
}```

BTW, is there a preferred avenue for me to seek assistance if I get stuck? I don't want to spam this issue.
For reference, the 2 versions of the files I'm currently trying to decode are:

[SYNT001.XML](https://github.com/transat/amarara/blob/main/SYNT001.XML)
[SYNT003.XML](https://github.com/transat/amarara/blob/main/SYNT003.XML)

MaxDesiatov:
Yes, I think your suggestion with one model type and branching with `xmlFileRootElementToString` could work, although I'm not sure how easy it will be to get that value passed to that function.

Please feel free to create a separate issue if you find any problems.

MaxDesiatov:
Hi @bwetherfield, would you be able to finish this up? If you're currently busy, I hope I can pick it up at some point.

bwetherfield:
Hi @MaxDesiatov. Sorry - forgot about this! I took a look and got _close_ to a satisfactory solution I think. I can't get `ExpressibleByNilLiteral` conformance working for now, which is kind of a deal breaker. That said, CI is all red (tests clear locally) - do you know what's going on there?

MaxDesiatov:
Thanks!

Ci failures are caused by SwiftLint, which complains about force casts. If those can't be avoided you can add something like `// swiftlint:dissable:next force_cast` on a line that preceeds a force cast.

bwetherfield:
OK, I gave that a try - I don't know of any other way to do `Expressible...` conformance, so I tried to enable the forcecasts. Still may be going red.

This is probably the most work I can do on this for the next while as I'm tied up in other projects. I'm happy to close or have you try to tackle some, @MaxDesiatov. Thanks and sorry for the limited bandwidth!

MaxDesiatov:
No problem at all, I can pick it up from here. Thanks for you work!

MaxDesiatov (approved):
👍

bwetherfield:
Great job with this @MaxDesiatov. I always learn a lot from reading over your code! Thanks for getting this PR over the line.

transat:
Are there any instructions for using this? Cheers!

MaxDesiatov:
We've discussed some basic use in https://github.com/MaxDesiatov/XMLCoder/issues/189, but since this feature is only available in the `main` branch, I haven't added any docs to `README.md` yet.

PR #200, opened 2020-08-30
Compile as dynamic library on Linux

colemancda:
Build as .so instead of .a on Linux for use across different libraries and executables.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,12 +3,19 @@
 
 import PackageDescription
 
+#if os(Linux)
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
Thanks for the PR. Could you clarify please how is this supposed to work if our Linux users want a static library? Last time I talked to SwiftPM folks, their opinion was that this should be controlled on library consumer side, not library vendor side, otherwise consumers lose all flexibility.

MaxDesiatov:
With no updates from the OP I'm closing this as stale.

PR #214, opened 2021-02-27
Add convert from UPPERCASE decoding key strategy

huwr:
This will convert from `ALL_CAPS_SNAKE_CASE` to `camelCase`.

Previously, there was only snake_case, kebab-case, and first-letter-only Capitalised. To do ALL_CAPS, one needed to use a custom key decoding strategy.

I've been using this to make a decoder for [BeerXML](http://www.beerxml.com), which tends to have the keys capitalised.

### DIFF
--- a/Tests/XMLCoderTests/Minimal/KeyedTests.swift
+++ b/Tests/XMLCoderTests/Minimal/KeyedTests.swift
@@ -129,6 +129,26 @@ class KeyedTests: XCTestCase {
         XCTAssertThrowsError(try decoder.decode(ContainerCamelCase.self, from: xmlData))
     }
 
+    func testConvertFromAllCaps() throws {
+        let decoder = XMLDecoder()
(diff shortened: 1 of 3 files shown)


Tests/XMLCoderTests/Minimal/KeyedTests.swift
```diff
@@ -129,6 +129,26 @@ class KeyedTests: XCTestCase {
...
+            </CONTAINER>
+            """
+        let xmlData = xmlString.data(using: .utf8)!
+
+        let decoded = try decoder.decode(ContainerCamelCase.self, from: xmlData)
+
+        XCTAssertEqual(decoded.valUe, ["foo": 12])
+    }
+
+
```
MaxDesiatov, line 151:
```suggestion
```

MaxDesiatov (approved):
Sorry for the delay. This looks great, thanks!

PR #221, opened 2021-07-27
chore: filter out pure whitespace with nested elements

wooj2:
Proof of concept to filter out pure whitespace elements that are siblings of elements that are not pure whitespace.


wooj2:
I conducted a benchmark test of this approach vs the approach proposed in #222 --they seem to be equivalent in terms of performance.  I think it makes sense to put this logic in the stack parser rather than KeyedStorage, so I'm closing this PR in favor of #222

PR #223, opened 2021-08-02
Encode element with empty key, empty element, and attributes

wooj2:
Original description is in this issue:
https://github.com/MaxDesiatov/XMLCoder/issues/224

### DIFF
--- a/Sources/XMLCoder/Auxiliaries/XMLCoderElement.swift
+++ b/Sources/XMLCoder/Auxiliaries/XMLCoderElement.swift
@@ -217,12 +217,12 @@ struct XMLCoderElement: Equatable {
         _ string: inout String,
         _ charactersEscapedInAttributes: [(String, String)]
     ) {
-        let actuallyAttributes = self.elements.filter {
-            $0.key.isEmpty && $0.elements.isEmpty && !$0.attributes.isEmpty
+        let attributesBelongingToContainer = self.elements.filter {
+            $0.key.isEmpty && !$0.attributes.isEmpty
         }.flatMap {
             $0.attributes
         }
-        let allAttributes = self.attributes + actuallyAttributes
+        let allAttributes = self.attributes + attributesBelongingToContainer
 
         let attributes = formatting.contains(.sortedKeys) ?
             allAttributes.sorted(by: { $0.key < $1.key }) :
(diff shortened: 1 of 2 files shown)


Sources/XMLCoder/Auxiliaries/XMLCoderElement.swift
```diff
@@ -217,9 +217,16 @@ struct XMLCoderElement: Equatable {
         _ string: inout String,
         _ charactersEscapedInAttributes: [(String, String)]
     ) {
+        let actuallyAttributes = self.elements.filter {
+            $0.key.isEmpty && $0.elements.isEmpty && !$0.attributes.isEmpty
```
MaxDesiatov, line 221: is `key` a critical check here? Maybe your original case fails because of the key value intrinsic feature described here? https://github.com/MaxDesiatov/XMLCoder#coding-key-value-intrinsic

MaxDesiatov (changes requested):
Would you be able to add some unit tests here? On the surface these changes do make sense if they resolve your use case. But I'd like to see that use case isolated as unit test first.

MaxDesiatov (approved):
Seems legit, thanks

Tests/XMLCoderTests/Auxiliary/XMLElementTests.swift
```diff
@@ -64,4 +64,68 @@ class XMLElementTests: XCTestCase {
...
+        let simpleScalarPropertiesInput = Element(key: "SimpleScalarPropertiesInput",
+                                                  elements: [nested],
+                                                  attributes: [simpleScalarPropertiesInputNamespace])
+
+        let result = simpleScalarPropertiesInput.toXMLString(
+            escapedCharacters: (elements: XMLEncoder().charactersEscapedInElements, attributes: XMLEncoder().charactersEscapedInAttributes),
+            formatting: [],
+            indentation: .spaces(4)
+        )
+
```
MaxDesiatov, line 92:
```suggestion
        // swiftlint:disable line_length
```

wooj2:
> Seems legit, thanks

WOW you are toooo quick! haha.. I removed the ignore swift lint, and just reduced the name so that it is under 200 characters, thanks!

PR #225, opened 2021-08-14
Remove unneeded parameter `key`

MartinP7r:
The parameter was not used inside the method and only had one call-site

Please ignore if you are planning to use the parameter in some way.

All tests are passing


MaxDesiatov (approved):
seems legit

MartinP7r:
Thank you!  

Any idea what the intention behind the parameter was in the first place?

MaxDesiatov:
Thanks for the contribution! I wouldn't remember at this point. It could've been introduced accidentally, or it could be something vestigial and lost its purpose after some refactoring. The main thing is that it's now cleaned up 🙂

PR #230, opened 2021-10-26
Update XMLCoderElement.swift

ThePredators:
The existing implementation was wrong, when the element is empty we set an empty xml tag. We should simply not set it. 

Example :
`
structure Test { var Title: String, var description: String? }
`

In this case calling encode() function like this : 
`
    let xmlEncoder = XMLEncoder()
    xmlEncoder.keyEncodingStrategy = XMLEncoder.KeyEncodingStrategy.capitalized
    let data = try xmlEncoder.encode(self, withRootKey:"Test")
`

Result in : 
`
<Test><Title>"Test"</Title><Description /></Test>
`

The update I made will simply return this instead of an xml string containing an empty element : 
`
<Test><Title>"Test"</Title></Test>
`

### DIFF
--- a/Sources/XMLCoder/Auxiliaries/XMLCoderElement.swift
+++ b/Sources/XMLCoder/Auxiliaries/XMLCoderElement.swift
@@ -278,12 +278,12 @@ struct XMLCoderElement: Equatable {
         }
         var string = prefix
 
-        if !key.isEmpty {
-            string += "<\(key)"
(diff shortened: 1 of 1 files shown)


MaxDesiatov:
Hi @ThePredators, would you be interested to move this PR forward? I'm cleaning up the repository, and if this PR is abandoned and outdated, I'm inclined to close it. Thanks!

MaxDesiatov:
I'm closing this PR as abandoned, please feel free to reopen otherwise.

PR #234, opened 2021-11-07
Add test for preserved spaces with xml entities

MartinP7r:
As mentioned in  https://github.com/MaxDesiatov/XMLCoder/pull/226#issuecomment-962542834 this test makes sure that strings containing xml entities surrounded with whitespaces are handled correctly.


MaxDesiatov (approved):
Thanks!

PR #242, opened 2022-07-17
Fix iOS 11 support warnings

kikeenrique:
Hi,
I would like to contribute fixing a recent warning.

The new Xcode 14 uses next SDKs as base: iOS 11.0, macOS 10.13, watchOS 4.0 and tvOS 11.0. Compiling in this Xcode shows a warning.
Handle this restrictions relying on swift version (which is also vinculated to each Xcode version).

Regards and comments welcomed

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -3,8 +3,15 @@
 
 import PackageDescription
 
+#if swift(>=5.7)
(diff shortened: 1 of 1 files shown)


MaxDesiatov (approved):
Seems legit, thanks!

PR #245, opened 2022-08-15
Test with Xcode 14.0 and Swift 5.7

MaxDesiatov:
The way the XML parser from Foundation treats errors differs between SDK versions, we have to handle this in our tests accordingly.

### DIFF
--- a/Tests/XMLCoderTests/AdvancedFeatures/ErrorContextTest.swift
+++ b/Tests/XMLCoderTests/AdvancedFeatures/ErrorContextTest.swift
@@ -31,26 +31,34 @@ final class ErrorContextTest: XCTestCase {
                 return
             }
 
-            #if os(Linux) && swift(<5.4)
-            // XML Parser returns a different column on Linux
-            // https://bugs.swift.org/browse/SR-11192
-            XCTAssertEqual(ctx.debugDescription, """
-            \(underlying.localizedDescription) \
-            at line 1, column 7:
-            `ah //>`
-            """)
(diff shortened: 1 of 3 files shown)


Tests/XMLCoderTests/AdvancedFeatures/ErrorContextTest.swift
```diff
@@ -31,26 +31,34 @@ final class ErrorContextTest: XCTestCase {
...
+            #elseif os(Windows) || os(Linux)
+            XCTAssertEqual(ctx.debugDescription, column10)
             #else
-            XCTAssertEqual(ctx.debugDescription, """
-            \(underlying.localizedDescription) \
-            at line 1, column 2:
-            `<blah `
-            """)
+            if #available(iOS 16.0, tvOS 16.0, macOS 13.0, *) {
+                XCTAssertEqual(ctx.debugDescription, column7)
```
Joannis, line 58: Do we (want to) guarantee error output from Foundation? If so, this PR alone seems to indicate we cannot.

MaxDesiatov: This output seemed to be pretty stable up to this point, and I haven't found a different way to test the error context functionality end-to-end without being exposed to these caveats with how Foundation handle this.

Joannis (approved):
I'm happy to merge this with 5.7 test support, but we do need to evaluate the errors at some point.

MaxDesiatov:
Thanks, I've created #248 to track this.

PR #246, opened 2022-08-17
Update doc comments, add `PropertyWrappersTest`

MaxDesiatov:
When viewing documentation published at https://swiftpackageindex.com/CoreOffice/XMLCoder/main/documentation/xmlcoder, I noticed that some of the doc comments are missing, and we made some of the protocols `public` by mistake. I've added doc comments with example code and update those protocols accordingly.

Additionally, tests and a section in `README.md` for recently added property wrappers were missing, that's fixed as well.

### DIFF
--- a/Sources/XMLCoder/Decoder/XMLDecoder.swift
+++ b/Sources/XMLCoder/Decoder/XMLDecoder.swift
@@ -39,7 +39,7 @@ open class XMLDecoder {
         static func keyFormatted(
             _ formatterForKey: @escaping (CodingKey) throws -> DateFormatter?
         ) -> XMLDecoder.DateDecodingStrategy {
-            return .custom { (decoder) -> Date in
+            return .custom { decoder -> Date in
                 guard let codingKey = decoder.codingPath.last else {
                     throw DecodingError.dataCorrupted(DecodingError.Context(
                         codingPath: decoder.codingPath,
@@ -90,7 +90,7 @@ open class XMLDecoder {
         static func keyFormatted(
             _ formatterForKey: @escaping (CodingKey) throws -> Data?
         ) -> XMLDecoder.DataDecodingStrategy {
-            return .custom { (decoder) -> Data in
+            return .custom { decoder -> Data in
                 guard let codingKey = decoder.codingPath.last else {
                     throw DecodingError.dataCorrupted(DecodingError.Context(
                         codingPath: decoder.codingPath,
@@ -179,7 +179,7 @@ open class XMLDecoder {
         }
 
         static func _convertFromUppercase(_ stringKey: String) -> String {
-          _convert(stringKey.lowercased(), usingSeparator: "_")
+            _convert(stringKey.lowercased(), usingSeparator: "_")
         }
 
         static func _convertFromSnakeCase(_ stringKey: String) -> String {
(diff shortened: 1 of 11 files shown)


MaxDesiatov:
Hi @CoreOffice/xmlcoder-team, would you have a moment to take a look? I wanted to make sure we're on the same page in terms of documentation, so your PR review here would be appreciated. Thanks!

Sources/XMLCoder/Decoder/XMLDecoder.swift
```diff
@@ -252,8 +252,12 @@ open class XMLDecoder {
 
     /// A node's decoding type
     public enum NodeDecoding {
+        /// Decodes a node from attributes of form `nodeName="value"`.
         case attribute
+        /// Decodes a node from elements of form `<nodeName>value</nodeName>`.
         case element
+        /// Decodes a node from either elements of form `<nodeName>value</nodeName>` or attributes
+        //// of form `nodeName="value"`.
```
Joannis, line 260: Do we guarantee the order in which the check is made?

Tests/XMLCoderTests/AdvancedFeatures/PropertyWrappersTest.swift
```diff
@@ -0,0 +1,69 @@
...
+        encoder.outputFormatting = .prettyPrinted
+
+        let xml = try String(data: encoder.encode(book), encoding: .utf8)
+
+        XCTAssertEqual(bookAuthorElementAndAttributeXML, xml)
+    }
+
+    func testDecode() throws {
+        let decoder = XMLDecoder()
+        let decodedBookBoth = try decoder.decode(Book.self, from: Data(bookAuthorElementAndAttributeXML.utf8))
```
Joannis, line 61: Shouldn’t we also add a from string decoding API, if one is seemingly lacking. That is not a change request, but a question

Joannis:
What are the odds that people currently rely on this API? I’m not exactly for a breaking change, but I wonder what the impact would be

Tests/XMLCoderTests/AdvancedFeatures/PropertyWrappersTest.swift
MaxDesiatov: I think this would be a sensible addition and raises a question why `JSONDecoder` haven't got anything like this by now? XMLCoder was modelled after `JSONDecoder` and `JSONEncoder` APIs.

MaxDesiatov:
> What are the odds that people currently rely on this API? I’m not exactly for a breaking change, but I wonder what the impact would be

If you're referring to `XMLAttributeProtocol`, I think the odds are vanishingly low. It didn't bring in any new functionality to users and is just an internal hack to type-erase values stored in property wrappers, exposed as `public` by accident.

In theory, I could expect someone conforming to this protocol similarly by accident, but even doing that would require some conscious effort on their side, and they shouldn't be doing this in the first place. Since XMLCoder 1.0 haven't been tagged yet so far, I think it's acceptable to introduce this breaking change for the sake of having a cleaner API.

Sources/XMLCoder/Decoder/XMLDecoder.swift
MaxDesiatov: Great catch, I've clarified it in the comment

Joannis (approved):
Considering the internal API is not interesting to implement, and we're pre 1.0, I'm happy removing it in a minor. The rest is great!

PR #261, opened 2023-03-15
Fix mixed CDATA usage

### LINKED ISSUE #260: CDATA usage in array of tags gives empty string instead of value
I am trying to get this failing test to work, but not getting anywhere.

```swift
import XCTest
import XMLCoder

final class CDATAEntryTest: XCTestCase {
    private struct DataEntry: Codable, Equatable {
        let value: String
            case value = ""
        }
    }
    
    private struct Container: Codable, Equatable {
        let data: [DataEntry]
    }
    
    private let xml =
        """
        <container>
           <data><![CDATA[lorem ipsum]]></data>
           <data>bla bla</data>
        </container>
        """.data(using: .utf8)!
    
    func testXMLWithCDataDataEntries() throws {
        let decoder = XMLDecoder()
        let result = try decoder.decode(Container.self, from: xml)
        XCTAssertEqual(result, Container(data: [.init(value: "lorem ipsum"), .init(value: "bla bla")])) // FAILS: got: ["", "bla bla"], but expected: ["lorem ipsum", "bla bla"]
    }
}
```

johankool:
Fixes #260

### DIFF
--- a/Sources/XMLCoder/Decoder/XMLKeyedDecodingContainer.swift
+++ b/Sources/XMLCoder/Decoder/XMLKeyedDecodingContainer.swift
@@ -275,9 +275,14 @@ extension XMLKeyedDecodingContainer {
         }
 
         // If we are looking at a coding key value intrinsic where the expected type is `String` and
-        // the value is empty, return `""`.
+        // the value is empty, return CDATA if present otherwise `""`.
         if strategy(key) != .attribute, elements.isEmpty, attributes.isEmpty, type == String.self, key.stringValue == "", let emptyString = "" as? T {
-            return emptyString
+            let cdata = container.withShared { keyedBox in
+                keyedBox.elements["#CDATA"].map {
+                    return ($0 as? KeyedBox)?.value ?? $0
+                }
+            }.first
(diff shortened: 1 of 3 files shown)


Joannis:
Thanks for your contribution! I'll let the tests run, and run it through myself. But it looks good for sure.

Joannis:
Done

johankool:
Not sure what to do about this. I don't have Xcode 13.1 on macOS 11.0 available to me, so can't easily reproduce. Not sure why anyone would still use that either.

Joannis:
I agree with you @johankool , it's an old platform. And even the Xcode build is old for the specific platform. If you want, you can replace the target with a Ventura / Xcode 14 one to get the tests running. I'll make sure we'll also add Linux/5.7 and Linux/5.8 after that.

Joannis:
Actually, the latest tests (Xcode 14, macOS 12) never ran to begin with. So I do think there's _something_ at play.

Joannis (approved):

Joannis:
I feel like this is a github runner issue, so I've merged it.

PR #264, opened 2023-04-18
Adds `.noEmptyElements` option to OutputFormatting

nighthawk:
Opts out of empty element shorthand.

---

I'm using XMLCoder to generate XML from my Swift data models for use in another system that doesn't handle the empty-elements, so I've added this option to allow generating XML that it can ingest. Not sure if `OutputFormatting` is the right place for this.

### DIFF
--- a/Tests/XMLCoderTests/NodeEncodingStrategyTests.swift
+++ b/Tests/XMLCoderTests/NodeEncodingStrategyTests.swift
@@ -313,11 +313,7 @@ final class NodeEncodingStrategyTests: XCTestCase {
 
     func testNoEmptyElements() {
         let encoder = XMLEncoder()
-        if #available(macOS 10.13, *) {
-            encoder.outputFormatting = [.noEmptyElements]
-        } else {
-            return
-        }
+        encoder.outputFormatting = [.noEmptyElements]
 
         do {
             let data = try encoder.encode(UnkeyedContainer(elements: []), withRootKey: "container")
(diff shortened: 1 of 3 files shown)


Tests/XMLCoderTests/NodeEncodingStrategyTests.swift
```diff
@@ -309,4 +309,27 @@ final class NodeEncodingStrategyTests: XCTestCase {
             XCTAssert(false, "failed to decode the example: \(error)")
         }
     }
+    
+
+    func testNoEmptyElements() {
+        let encoder = XMLEncoder()
+        if #available(macOS 10.13, *) {
```
Joannis, line 316: Why is this wrapped in `#available(macOS 10.13, *)`?

Joannis:
I think that OutputFormatting is fine. Thanks for the contribution! I think I can merge this if the tests succeed

Joannis:
I know that the one test that failed in your previous commit is not your fault.

Tests/XMLCoderTests/NodeEncodingStrategyTests.swift
nighthawk: Good question. I started with the previous test as a base, which had the same check around it. I've removed it.

Note that all tests in that class have the same check which dates back several years, though my Xcode is fine when I remove them. Maybe it's not be needed anymore due to this in the Package.swift file:

```
#if swift(>=5.7)
let platforms: [PackageDescription.SupportedPlatform] = [.macOS(.v10_13), .iOS(.v11), .watchOS(.v4), .tvOS(.v11)]
#elseif swift(>=5.0)
let platforms: [PackageDescription.SupportedPlatform]? = nil
#endif
```

Though not sure what'll happen in Swift versions earlier than 5.7.

Joannis:
Tagged as 0.17.0

nighthawk:
Thank you, @Joannis!

PR #265, opened 2023-04-26
Fix decoding of enum with associated values when enum case does not contain value

### LINKED ISSUE #202: How to encode empty tags as enum based on name
Hey, at first thanks a lot for making this library, great work! 

I am rather new to the XML game and trying to decode WebDAV responses. 
So I have the following response from the server: 

```xml
<?xml version="1.0" standalone="yes"?>
<d:multistatus xmlns:d="DAV:" xmlns:s="http://sabredav.org/ns">
    <d:response>
        <d:href>someURL</d:href>
        <d:propstat>
            <d:prop>
                <d:resourcetype>
                    <d:collection></d:collection>
                </d:resourcetype>
            </d:prop>
            <d:status>HTTP/1.1 200 OK</d:status>
        </d:propstat>
    </d:response>
</d:multistatus>
```

And the following `Codable` models: 

```swift
struct MultiStatus: Codable {
    let responses: [MultiStatusResponse]
    
        case responses = "response"
    }
}

struct MultiStatusResponse: Codable {
    struct Property: Codable {
        enum ResourceType: Codable {
            case collection
            case file
            case redirect
        }
        
        let ressourceType: ResourceType
    }

    let href: String
    let properties: [Property]
    let error: String?
    
        case href
        case properties = "prop"
        case error
    }
}
```
-> This implementation does not work, because it is complaining about the `ResourceType` enum not conforming to `Codable` because of the missing  method implementations of `decode` and `encode`.

I am searching for a way to decode the 
```xml
  <d:resourcetype>
      <d:collection></d:collection>
  </d:resourcetype>
```
part into an enum case based on the tag inside the `<d:resourcetype>` tag. So in this case I would like the `ResourceType` enum to be of case `collection`. Is that possible with your library or do I need to go one level deeper and parse the XML by myself ? I tried to write custom `decode` and `encode` methods but unfortunately didn't get it to work.

Alkenso:
Consider enum with associated values, where some value is empty
```
enum Choice: Equatable, Codable {
      case one
      case two(Int)
      case three(String, value: Int)
}
```

This PR fix decoding of such enum (while encoding is already worked as expected)
See new test in commit for details

Seems it also fixes #202

### DIFF
--- a/Tests/XMLCoderTests/AdvancedFeatures/SimpleChoiceTests.swift
+++ b/Tests/XMLCoderTests/AdvancedFeatures/SimpleChoiceTests.swift
@@ -85,4 +85,58 @@ final class SimpleChoiceTests: XCTestCase {
         let decoded = try XMLDecoder().decode([[IntOrString]].self, from: encoded)
         XCTAssertEqual(original, decoded)
     }
+    
+#if swift(>=5.5)
+    func testMixedChoice() throws {
+        enum Choice: Equatable, Codable {
+            case one
+            case two(value: Int)
+            case three(key: String, value: Int)
+            case four(Int)
+        }
+
+        struct Foo: Equatable, Codable {
+            var field1 = Choice.one
+            var field2 = Choice.two(value: 10)
+            var field3 = Choice.three(key: "qq", value: 20)
+            var field4 = Choice.four(30)
+        }
+        
+        let encoder = XMLEncoder()
+        encoder.outputFormatting = .prettyPrinted
+        encoder.prettyPrintIndentation = .spaces(4)
+        
+        let original = Foo()
+        let encoded = try encoder.encode(original, withRootKey: "container")
+        let decoded = try XMLDecoder().decode(Foo.self, from: encoded)
+        XCTAssertEqual(original, decoded)
+        
+        XCTAssertEqual(
+            String(data: encoded, encoding: .utf8),
+            """
+            <container>
+                <field1>
+                    <one />
+                </field1>
+                <field2>
+                    <two>
+                        <value>10</value>
+                    </two>
+                </field2>
+                <field3>
+                    <three>
+                        <key>qq</key>
+                        <value>20</value>
+                    </three>
+                </field3>
+                <field4>
+                    <four>
+                        <_0>30</_0>
+                    </four>
+                </field4>
+            </container>
+            """
+        )
+    }
+#endif
 }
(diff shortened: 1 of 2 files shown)


Alkenso:
@johankool , please comment or approve the PR

MaxDesiatov:
I no longer maintain this library, please avoid tagging me in PRs or issues I don't work on. Thanks!

Alkenso:
@Joannis , maybe you could take a look?

FYI: some tests are broken because of lack Swift support for codable enums with associated values in old versions of Xcode

Tests/XMLCoderTests/AdvancedFeatures/SimpleChoiceTests.swift
```diff
@@ -85,4 +85,23 @@ final class SimpleChoiceTests: XCTestCase {
         let decoded = try XMLDecoder().decode([[IntOrString]].self, from: encoded)
         XCTAssertEqual(original, decoded)
     }
+    
+    func testMixedChoice() throws {
+        enum Choice: Equatable, Codable {
```
Joannis, line 90: I think that the issue is that Xcode 11.x can't yet generate the codable implementation for this type.
We can wrap this test in a compiler directive like `#if compiler` or `#if swift`.

Tests/XMLCoderTests/AdvancedFeatures/SimpleChoiceTests.swift
```diff
@@ -85,4 +85,23 @@ final class SimpleChoiceTests: XCTestCase {
...
+
+        struct Foo: Equatable, Codable {
+            var one = Choice.one
+            var two = Choice.two(10)
+            var three = Choice.three("qq", value: 20)
+        }
+        
+        let original = Foo()
+        let encoded = try XMLEncoder().encode(original, withRootKey: "container")
+        let decoded = try XMLDecoder().decode(Foo.self, from: encoded)
```
Joannis, line 104: If possible, I'd love a simple test that demonstrates the expected XML output.

Joannis:
Thanks for pinging me @Alkenso , feel free to ping me anytime I overlook a PR or issue.
It's impossible to support a Swift feature that that compiler version itself does not yet support, so you get a free pass there with a compiler directive to let the tests go on.

Tests/XMLCoderTests/AdvancedFeatures/SimpleChoiceTests.swift
Alkenso: done

Tests/XMLCoderTests/AdvancedFeatures/SimpleChoiceTests.swift
Alkenso: added output test

Alkenso:
@Joannis  Please re-run failed job

Alkenso:
Looks very strange. Especially that all works fine on newer and older versions.
Maybe something broken on CI? How can I check it?

Joannis:
@Alkenso a previous PR also failed on the same runner, I think this is a CI failure on Xcode 13.1/macOS 11.

Joannis (approved):
Looks good!

Issue #254, opened 2022-12-07
DynamicNodeDecoding cause producing incorrect error on failure in nested struct


Alkenso:
The issue makes me crazy when I'm looking error logs. Probably easy to fix, but will help a lot understanding what's went wrong.

```
// Assume we have XML:
let xml =
"""
<Root>
   <field attr1="value_1" />
</Root>
"""

// And want to decode it into struct:
struct Model: Decodable {
    var field: Field
    
    struct Field: Decodable, DynamicNodeDecoding {
        var attr1: String
        var attr2: String
        
        static func nodeDecoding(for key: CodingKey) -> XMLDecoder.NodeDecoding {
            .attribute
        }
    }
}

// Using XMLDecoder:
do {
    _ = try XMLDecoder().decode(Model.self, from: Data(xml.utf8))
} catch {
    print(error)
}
```

**Actual**
The error tells that whole property `field` is null
```
▿ DecodingError
  ▿ valueNotFound : 2 elements
    ▿ .1 : Context
      ▿ codingPath : 1 element
        - 0 : CodingKeys(stringValue: "field", intValue: nil)
      - debugDescription : "Expected Field value but found null instead."
      - underlyingError : nil
```

**Expected**
If we remove conformance to `DynamicNodeDecoding`, we get correct error
```
▿ DecodingError
  ▿ keyNotFound : 2 elements
    - .0 : CodingKeys(stringValue: "attr2", intValue: nil)
    ▿ .1 : Context
      ▿ codingPath : 2 elements
        - 0 : CodingKeys(stringValue: "field", intValue: nil)
        - 1 : CodingKeys(stringValue: "attr2", intValue: nil)
      - debugDescription : "No attribute or element found for key CodingKeys(stringValue: \"attr2\", intValue: nil) (\"attr2\")."
      - underlyingError : nil
```

Joannis:
Thanks for the bug report! I don't have time the coming week, but will make put it on my agenda after that.

Alkenso:
Fixed

PR #266, opened 2023-05-09
Fix decoding of types conforming to XMLNodeDecoding

Alkenso:
Fix case when XMLNodeDecoding wants to decode elementOrAttribute, but value is not present in XML
Was: silently treat missing of XML child as default value for string
Now: throws error

Fix error reporting when key not found for element or attribute #254


Alkenso:
FYI: @Joannis

Joannis:
I've added some extra tests so that it covers unkeyed containers too

